PR #14, opened 2014-11-05
Allow own FB APP ID to be deployed

fgrosswig:
plugin.tx_news.settings.facebookAppId = yourownappid


georgringer:
I don't plan to invest more time in the sovial VH. Use ext "rx_shariff" instead. therefore I am closing this issue

PR #20, opened 2015-01-26
[BUGFIX] Clear template path befor fallback is set

misterboe:


georgringer:
there are 2 use cases:
- either someone doesn't include the default TS and uses his own code
- someone includes the default TS as base and includes his changes afterwards

both uses cases are supported and valid but for sure there are many others which might not lead to a desired behaviour but I can't support them all. therefore I am closing this PR

PR #24, opened 2015-04-22
[BUGFIX] Fix empty field selector in update script

coding46:


georgringer:
this has been fixed in the meantime, sorry for late response!

PR #26, opened 2015-06-02
Fix typo in locallang_db.xlf

inrivagmbh:


georgringer:
this has been fixed in the meantime. sorry that I have seen that so late but developement happens in gerrit!

inrivagmbh:
thanks :)

PR #44, opened 2015-11-10
[BUGFIX] Import localized subcategories

DanielSiepmann:
- Process import for subcategories even if they are localized

### DIFF
--- a/Classes/Domain/Service/CategoryImportService.php
+++ b/Classes/Domain/Service/CategoryImportService.php
@@ -80,43 +80,48 @@ class CategoryImportService extends AbstractImportService
     {
         $this->logger->info(sprintf('Starting import for %s categories', count($importArray)));
 
-        // Sort import array to import the default language first
         foreach ($importArray as $importItem) {
+            $postPersistQueueItem = array();
             $category = $this->hydrateCategory($importItem);
 
             if (!empty($importItem['title_lang_ol'])) {
-                $this->postPersistQueue[$importItem['import_id']] = [
+                $postPersistQueueItem = array(
                     'category' => $category,
                     'importItem' => $importItem,
-                    'action' => self::ACTION_CREATE_L10N_CHILDREN_CATEGORY,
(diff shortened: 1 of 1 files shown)


Classes/Domain/Service/CategoryImportService.php
```diff
@@ -80,43 +80,48 @@ public function import(array $importArray)
...
+                );
             }
 
             if ($importItem['parentcategory']) {
-                $this->postPersistQueue[$importItem['import_id']] = [
-                    'category' => $category,
-                    'action' => self::ACTION_SET_PARENT_CATEGORY,
-                    'parentCategoryOriginUid' => $importItem['parentcategory']
-                ];
+                $postPersistQueueItem = \TYPO3\CMS\Core\Utility\GeneralUtility::array_merge(
```
NamelessCoder: Alternatively: `$postPersistQueueItem += ['category' => $category, 'parentCategoryOriginUid' => $importItem['parentCategory']];`

Classes/Domain/Service/CategoryImportService.php
```diff
@@ -80,43 +80,48 @@ public function import(array $importArray)
...
+            $postPersistQueueItem = array();
             $category = $this->hydrateCategory($importItem);
 
             if (!empty($importItem['title_lang_ol'])) {
-                $this->postPersistQueue[$importItem['import_id']] = [
+                $postPersistQueueItem = array(
                     'category' => $category,
                     'importItem' => $importItem,
-                    'action' => self::ACTION_CREATE_L10N_CHILDREN_CATEGORY,
+                    'action' => array(self::ACTION_CREATE_L10N_CHILDREN_CATEGORY),
```
NamelessCoder: Rather than turn this into an array I would simply create a copy of the $postPersistQueueItem with a different action, when this is needed. This yields two queued unique actions which seems cleaner - and it means this diff can be much smaller / change less business logic.

DanielSiepmann:
I'm no longer into the code. The PR was created over a year ago. This way I can't adjust the Pull Request anymore, sorry.

Just close it and reopen once someone else has the same issue.

PR #60, opened 2016-03-11
fix non visible translated categories in filter

kanow:


georgringer:
fixed with https://review.typo3.org/#/c/46871/ already

PR #65, opened 2016-04-06
Fixes #64

### LINKED ISSUE #64: use initializeObject instead of __construct in news model
Usage of __construct leads to this behaviour:

http://lists.typo3.org/pipermail/typo3-german/2014-August/104462.html

Using initializeObject instead of __construct fixes it.

contemas-tschmidt:
Use initializeObject instead of __construct in domain model news


georgringer:
I never had such problems, therefore I am closing this issue

bernhardberger:
Sorry for reopening this.. but this actually is a real problem. 

If you instantiate classes via Extbase the constructor doesn't get called in some cases as the object is created by some reflection magic rather than instantiated by the 'new' keyword.

Maybe you don't encounter that as much (or at all) as you most likely very seldomly extend your own extension..

PR #72, opened 2016-05-03
[FEATURE] add index for tt_content::tx_news_related_news

franzkugelmann:
we experienced quite some load on the mysql-Server with queries fetching the related content for news.
We have > 100.000 entries in tt_content.
Also for smaller sites i see no harm in having such an index.

EXT:news rocks! Keep going!


georgringer:
thank  for your PR! next time please use gerrit as it makes it easier for me! merged it with https://review.typo3.org/#/c/48075/

PR #84, opened 2016-07-13
link tags and categories in Detail-View

frankschmelcher:
Linking tags and categories in Detail-View to the Back-Link-Page works fine with this change.
If there is no Back-Link defined, the tags and categories are not linked --> fine


georgringer:
linking tags and categories to the backpid works for your usecase but is not obvious for me as default behaviour. as this change is only needed for templates and those can be easily adopted, I won't merge this.

PR #88, opened 2016-08-11
add fix for language strict mode

phe-move-elevator-de:
Images from news objects are not displayed if language not default and language strict mode is active. The problem occurs because FAL are not saved with the correct sys_language_uid from the parent object. All FAL objects are saved with the sys_language_uid 0. To solve this Problem, there must be a possibility to set the language for the FAL object.


georgringer:
this still doesn't fix a lot other issues like if an image is only added in the translation

tmotyl:
related: #260

PR #96, opened 2016-09-13
Fix wrong insertion of fe_group on repeated import

sascha-schieferdecker:
Without the if condition, the string 'NULL' is inserted into the corresponding field. The imported news aren't displayed anymore, because an access restriction is in effect. This doesn't happen on first import, but if the news record already exists.


Classes/Domain/Service/NewsImportService.php
```diff
@@ -151,7 +151,11 @@ protected function hydrateNewsRecord(
         $news->setHidden($importItem['hidden']);
         $news->setStarttime($importItem['starttime']);
         $news->setEndtime($importItem['endtime']);
-        $news->setFeGroup((string)$importItem['fe_group']);
+        
+        if(!empty($importItem['fe_group'])) {
+            $news->setFeGroup($importItem['fe_group']);
```
georgringer: can you please still force the (string)?

PR #100, opened 2016-09-15
[BUGFIX] Importer broken

geraldaistleitner:
Importer was broken as the change of the backend module position to system was not taken care of in templates.


georgringer:
thanks! merged the gerrit patch

PR #110, opened 2016-09-26
Fix wrong insertion of fe_group on repeated import

sascha-schieferdecker:
...now with typecasting...


georgringer:
I was not able to merge that, therefore I commited it as new patch. thanks for your work!

Issue #117, opened 2016-10-02
Exploit new Fluid capabilities (TYPO3v8+)


NamelessCoder:
This issue is to both discuss and demonstrate how new Fluid features (since Fluid standalone) can be utilised in EXT:news to drastically improve ViewHelper performance.

See commit and diff of current WIP: https://github.com/georgringer/news/compare/master...NamelessCoder:newfluidviewhelpers?expand=1

Reference: https://github.com/TYPO3Fluid/Fluid/commit/1f50ae30846c9deb6a0ed14f913f558e41b9e9f8

**Note! Impossible to achieve for now using current WIP - until TYPO3 7.6 is no longer supported in EXT:news! However, replicating the Traits and ViewHelperCompiler, adjusting the signatures along the way, would allow this to work on 7.6 as well.**

georgringer:
It would be great if you could split it up to make it possible to profit from changes already in 7.6. All those reflection changes could go in now already, right?

NamelessCoder:
If you are willing to have some temporary code duplication we can probably gain the full improvement now already. I will prepare a pull request to convert one of the ViewHelpers completely so you can see the intention behind it :)

georgringer:
@NamelessCoder should I leave this open?

NamelessCoder:
No need - I'll create new PRs as I go along with the static/compiling work!

PR #119, opened 2016-10-03
[TASK] Make ViewHelpers static callable and compilable

NamelessCoder:


NamelessCoder:
This PR shows the amount of code duplication we would require in order to achieve fully static calling and support for compiling (on any TYPO3 version as far as I can predict right now). You may prefer to have these duplicated methods on a Trait instead.

The one ViewHelper demonstrated here did not require render method argument migration but for obvious reasons, that operation would be required for all ViewHelpers which call any non-static member method.

NamelessCoder:
> as far as I can predict right now

Forgot about the signature change in `compile()` method combined with PHP 7+. Looks like the only solution for that part would be having two versions of the same-named Trait and place it in a dependency which we release in two versions compatible with 7.6 and 8+ respectively. I don't know if that's something to consider? Or maybe you know another way to make composer load an alternative class file based on TYPO3 version so we could ship both in EXT:news directly?

NB: the TYPO3 8 specific version of the Trait can be nothing but a class alias of the Trait from https://github.com/georgringer/news/issues/117#issuecomment-251093140 (when TYPO3 raises Fluid dependency to 1.1.0 which will be released very soon).

NamelessCoder:
@georgringer You can put this one temporarily on hold. I will be creating the mentioned dependency containing the trait with versions for TYPO3 7.6 and 8+. After thinking more about this, VHS as well as Flux would benefit from such a library while we wait for 7.6 support to expire. At that time we then simply remove the dependency and change our imported Trait to the final Fluid one.

Then I hope you will consider living with such a dependency for a little while in EXT:news... but don't decide until you see the result ;)

NamelessCoder:
I'm not sure if I did something wrong, but Travis is skipping the `TYPO3_VERSION=^8` builds for some reason. Aside from that: this is now ready and full compiling support and removal of duplicated render() methods can be achieved by merely raising the TYPO3 version requirement to ^7.6.13.

NamelessCoder:
Sorry about the ping, @georgringer - if you've got a moment, would you review this change and tell if anything more is required / if the minimum version change is acceptable? If this is okay I will proceed with other ViewHelpers! I will have some downtime on Sunday that I plan to spend some of on EXT:news.

georgringer:
I updated the travis now, so that should work fine. I hope I find some time now to check your changes!

composer.json
```diff
@@ -20,7 +20,7 @@
     "GPL-2.0+"
   ],
   "require": {
-    "typo3/cms-core": ">=7.6.0,<8.9.99"
+    "typo3/cms": ">=7.6.13,<8.9.99"
```
georgringer: Isn't that too simple as it would break than with 8.1 and all 8x versions?

georgringer: I am absolutly fine to force 8x users to latest version. So what about `"typo3/cms-core": ">=7.6.13 || >=8.5.0 <8.9.99"`

Would like to use the `-core`, following composer.json of typo3_console

NamelessCoder: Sounds fair enough to me - although it isn't technically necessary. Would you prefer that as a second commit we can merge first, then remove the composer change part of this PR?

On the subject of `-core` or not, I will also suggest to Helmut to use the `typo3-cms` name in `typo3_console` as well. Not to be disruptive, I just think it's the more correct way (given the current state of the TYPO3 composer manifest which `replaces` the `-core` name).

georgringer: I updated the composer.json already as separate commit ;)

NamelessCoder:
Looks cleaner now, should be ready :)

composer.json
helhum: @NamelessCoder 
Here is my take on that: 
http://insight.helhum.io/post/155297666635/typo3-extension-dependencies-revisited

PR #121, opened 2016-10-04
[BUGFIX] remove one row class from list container

### LINKED ISSUE #120: Bootstrap list view does not align with remaining rows
In the TWBS List.html template one row is nested in another row so the negative margin is doubled and so the list does not correctly align with other content on the page.

schugabe:
One row is nested in another row so the negative margin is doubled and so the list does not correctly align with other content on the page.
The outer row is removed since it's not necessary here.

Resolves #120

### DIFF
--- a/Resources/Private/Templates/Styles/Twb/Templates/News/List.html
+++ b/Resources/Private/Templates/Styles/Twb/Templates/News/List.html
@@ -11,7 +11,7 @@
 	<!--TYPO3SEARCH_end-->
 	<f:if condition="{news}">
 		<f:then>
-			<div class="row news-list-view">
+			<div class="news-list-view">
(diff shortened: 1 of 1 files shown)


georgringer:
@schugabe shoudn't be the whole div `<div class="row news-list-item">` be removed from the *partial*? the row in the list is IMO correct

schugabe:
I would not remove the row from the item itself because the row ensures that no strange things happen when the content is collapsed on smaller devices (probably no problem with the simple col-md-3 col-md-9 grid but it can happen if more complicated column combinations are used).

If the outer row is needed to ensure that at least one row wraps the content I'd insert a col-xs-12 to wrap all news items in order to fix the margin of the outer row.

PR #129, opened 2016-10-14
[TASK] enable custom values for dateField

martinhummer:


georgringer:
because of security I won't change this back. Use a custom `demandClass` to change the behaviour

PR #146, opened 2016-10-30
[BUGFIX] Add correct content-type header in CMS 8

liayn:


georgringer:
@liayn how does this change behave in 7?

liayn:
I just found out that the old one-line syntax is deprecated in CMS 7 as well. So I'll push a new PR, getting it now right.

PR #163, opened 2016-11-14
[TASK] Changes how showinpreview is added to palette,

tomasnorre:
It eases the possibility to extend the newsPalette


tomasnorre:
Would be nice if it could be backported to the 6x branch.. if not i'll fork and do it myself :)

PR #172, opened 2016-11-24
[FEATURE] Add TypoScript setting to enable/disable opengraph tags

### LINKED ISSUE #165: Deactivate rendering of open graph tags for detail view
Maybe it's possible to add an TS setting for de/activating the OG tags in the detail view.

t3easy:
Resolves #165
@wowawebdesign Wanna review?


t3easy:
What should be moved?
Description, keywords? Or also twitter:* and fb:admins?

If i split that, it would lead to a breaking change, as someone could have overridden the partial to exclude e.g. keywords.

georgringer:
yes would be breaking.
the other way would be to use the new condition in the partial itself and just move it arount all social share stuff, yes including twitter & co. therefore I would rename it also to something different like shareTagsInHead

t3easy:
@georgringer I now refactored this.
I think this is the best non breaking solution atm.
Later on the partial should be renamed.

Configuration/TypoScript/setup.txt
```diff
@@ -185,6 +185,12 @@ plugin.tx_news {
 				creator = {$plugin.tx_news.opengraph.twitter.creator}
 			}
 		}
+		metadata {
+			keywords = 1
+			description = 1
+			opengraph = 1
+			twitter =< plugin.tx_news.settings.opengraph.twitter.site
```
NamelessCoder: I would switch this to reference the constant instead.

Resources/Private/Partials/Detail/Opengraph.html
```diff
@@ -1,46 +1,57 @@
 <html xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers"
 	  xmlns:n="http://typo3.org/ns/GeorgRinger/News/ViewHelpers"
 	  data-namespace-typo3-fluid="true">
-
-<f:if condition="{newsItem.alternativeTitle}">
-	<f:then><n:metaTag property="og:title" content="{newsItem.alternativeTitle}" /></f:then>
-	<f:else><n:metaTag property="og:title" content="{newsItem.title}" /></f:else>
+<f:comment><!--@todo: Rename partial to e.g. "Metadata" at next major release because this would be breaking--></f:comment>
```
NamelessCoder: I don't know exactly what Georg's preference is here, but for my own opinion: please avoid TODOs in the code. If there should be a rename of this template file, please create a pull request after this is merged (and make a note that merging that pull request constitutes a breaking change).

georgringer: yeah please no todos

PR #186, opened 2016-12-13
[TASK] Set field type exclude on true

ruudsilvrants:
It is desired to exclude the field type for editors when only using one type.


georgringer:
If you don't need the field, you can also hide those also with TsConfig or by changing the  TCA in your custom sitepackage. I don't plan to change this in the ext itself. Hope you understand

PR #214, opened 2017-01-14
[BUGFIX] Fixed error handling for redirectToListView configuration

### LINKED ISSUE #206: Bug using the detail page error handling "redirectToListView"?
Hey Georg,

I'm not quite sure if this is a bug or I am wrong...

I experienced the problem with a customer, that many website users run into 404s as the news were deleted or similar. Now the customer would like to redirect the users to the list view if the news can't be displayed.
I found your error handling feature. Thanks, just what I was looking for!

But, if I define the error handling to use "redirectToListView" configuration, I'm running into a 303 loop.
Somehow it ignores the action parameter in the url which after the first redirect is "list" so that it keeps calling the detail action and redirecting to list. Maybe a bug or maybe some missconfiguration in the setup of this customer.

However, wouldn't it be better to just pass the whole settings array to the "handleNoNewsFoundError" function and then redirect to the list page defined as "backPid"? Then we also would avoid possible duplicate content, as google could call a deleted news and then get the list view on a diferent url which would be duplicate content, I think.

I've created a patch. May you want to take a look at it?
And would you backport this also to version 3, if you accept the patch? This customer is still on 6.2.

Thanks in advance.

Best,
Alex

[_BUGFIX_303_loop_redirectToListView.zip](https://github.com/georgringer/news/files/693596/_BUGFIX_303_loop_redirectToListView.zip)

NamelessCoder:
As this configuration was generating a 303 loop, I've changed the function to pass the hole settings array and use the backPid setting to redirect to the list page defined in the plugin or TypoScript.

Resolves #206

### DIFF
--- a/Classes/Controller/NewsBaseController.php
+++ b/Classes/Controller/NewsBaseController.php
@@ -85,13 +85,13 @@ class NewsBaseController extends \TYPO3\CMS\Extbase\Mvc\Controller\ActionControl
 
         switch ($configuration[0]) {
             case 'redirectToListView':
-				$this->uriBuilder->reset();
-				$this->uriBuilder->setTargetPageUid($settings['backPid']);
-				$this->uriBuilder->setCreateAbsoluteUri(TRUE);
-				if (GeneralUtility::getIndpEnv('TYPO3_SSL')) {
-					$this->uriBuilder->setAbsoluteUriScheme('https');
-				}
-				$url = $this->uriBuilder->build();
+                $this->uriBuilder->reset();
+                $this->uriBuilder->setTargetPageUid($settings['backPid']);
+                $this->uriBuilder->setCreateAbsoluteUri(true);
+                if (GeneralUtility::getIndpEnv('TYPO3_SSL')) {
+                    $this->uriBuilder->setAbsoluteUriScheme('https');
+                }
+                $url = $this->uriBuilder->build();
 
                 $this->redirectToUri($url);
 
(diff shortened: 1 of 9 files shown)


Tests/Unit/Controller/NewsBaseControllerTest.php
```diff
@@ -61,7 +51,7 @@ public function NoNewsFoundConfigurationRedirectsToListView()
             ['redirect']);
         $mock->expects($this->once())
             ->method('redirect')->with('list');
-        $mock->_call('handleNoNewsFoundError', 'redirectToListView');
+        $mock->_call('handleNoNewsFoundError', array('detail' => array('errorHandling' => 'redirectToListView')));
```
NamelessCoder: This and all following instances of `array()` should be switched to shorthand syntax `[]`.

Classes/Controller/NewsBaseController.php
```diff
@@ -71,21 +71,30 @@ private function handleKnownExceptionsElseThrowAgain(\Exception $exception)
...
             return null;
         }
 
-        $configuration = GeneralUtility::trimExplode(',', $configuration, true);
+        $configuration = GeneralUtility::trimExplode(',', $settings['detail']['errorHandling'], true);
 
         switch ($configuration[0]) {
             case 'redirectToListView':
-                $this->redirect('list');
+				$this->uriBuilder->reset();
```
NamelessCoder: Indentation must be with spaces.

Issue #218, opened 2017-01-18
TYPO3 extension "newspreview"


wowawebdesign:
I found the extension **newsinpreview**, which implements a very cool feature for displaying the preview images.
https://github.com/georgringer/TYPO3.CMS.newspreview

Would it be possible to integrate this as a core feature in news?

I know someone, who is willing to do the coding (@t3easy) ;)

georgringer:
should be doable

t3easy:
I've prepared a feature branch:
https://github.com/t3see/news/tree/feature-218
Feel free to test.

t3easy:
Hi @georgringer can you have a look at the branch to estimate if the implementation goes into the right direction?

georgringer:
Looks good from mobile. Some notes: check out the new getters in master without the fal in the name and use array short hand syntax

georgringer:
solved

PR #238, opened 2017-01-29
[FEATURE] Change showinpreview to preview mode

t3easy:
To allow to show a media only in preview/list the bool showinpreview was changed to a integer mode select.
It's extendable from outside to easy implement a new view mode, e.g. `show on home page` by adding a values to the TCA, extending the news class and add own getter and/or override existing ones.
Resolves: #218


georgringer:
would be great if you could take a look at the conflict. I will try to review + merge shortly after that then

r4fx:
Love to see it in next release

t3easy:
Hi @georgringer should I take a look at this feature afresh?

t3easy:
@r4fx Would you review, test and give feedback?

georgringer:
now pushed with some different code, thx for big inspiration

PR #262, opened 2017-02-21
[BUGFIX] Show create new tag at first position in results

liayn:


Classes/Hooks/SuggestReceiver.php
```diff
@@ -68,6 +68,13 @@ public function queryTable(&$params, $recursionCounter = 0)
                 'sprite' => '',
                 'icon' => $this->getDummyIcon()->render()
             ];
+
+            if (GeneralUtility::compat_version('8.0')) {
```
georgringer: this method is deprecated, please don't use it

Issue #271, opened 2017-03-02
[BUG] Utility\Page::extendPidListByChildren() returns wrong data for startingpoint = -1


georg-tiefenbrunn:
Hi Georg!

I'm struggling with news preview and workspaces in a multi domain/sites setup using TYPO3 7.6/EXT:news (composer).

Page structure:

```
TYPO3 (uid=0)
  + root (uid=1)
    + global records (uid=2)
      + news (uid=7)
        + website.a (uid=8)
          + department.1 (uid=11)
          + department.2 (uid=12)
          + department.3 (uid=13)
        + website.b (uid=9)
        + website.c (uid=10)
    + sites (uid=3)
        + website.a (uid=4)
        + website.b (uid=5)
        + website.c (uid=6)
```

All news plugins are rendered via TypoScript.

```typoscript
lib.newsPreview < lib.newsDetail
lib.newsPreview.10.settings {
    # Do not restrict categories in preview
    categories =
    # We use 'startingpoint = 0' here to disable public news preview.
    # The correct startingpoint is later set if a BE user is logged in.
    startingpoint = 0
    recursive = 0
}
[globalVar = TSFE:beUserLogin > 0]
    lib.newsPreview.10.settings {
        # Add '-1' to enable preview in workspaces
        startingpoint = -1,{$plugin.tx_news.settings.storagePid}
        recursive = {$plugin.tx_news.settings.recursive}
        enablePreviewOfHiddenRecords = 1
    }
[global]
```

*website.a* has ``plugin.tx_news.settings.storagePid = 8`` and ``plugin.tx_news.settings.recursive = 1`` to fetch all items from all departments.

All works fine in "LIVE" workspace. But I am experiencing some really strange behavior when trying  *news preview* in *workspace preview* (using "save and view" on a news item in a workspace).

To enable the preview of versioned news items I had to add ``-1`` to ``plugin.tx_news.settings.startingpoint``. For *website.a* the TS above results in  ``plugin.tx_news.settings.startingpoint = -1,8``.

As I don't want the news preview to show items from other folders ``plugin.tx_news.settings.detail.checkPidOfNewsRecord = true`` and so  [Controller\NewsController->checkPidOfNewsRecord()](https://github.com/georgringer/news/blob/master/Classes/Controller/NewsController.php#L285) calls [Page::extendPidListByChildren()](https://github.com/georgringer/news/blob/master/Classes/Utility/Page.php#L32) which returns wrong data for ``pid = -1``.

[...]

PR #272, opened 2017-03-02
[BUGFIX] Don't extend pid list for startingpoint=-1

georg-tiefenbrunn:
Using preview of single news item from workspaces might require to add '-1' to 'plugin.tx_news.settings.startingpoint'.
This fix ensures that no wrong child uids are fetched for 'pid = -1'.

Resolves: #271

### DIFF
--- a/Classes/Utility/Page.php
+++ b/Classes/Utility/Page.php
@@ -36,16 +36,19 @@ class Page
             return $pidList;
         }
 
+        /** @var \TYPO3\CMS\Core\Database\QueryGenerator $queryGenerator */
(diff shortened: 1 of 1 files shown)


Classes/Utility/Page.php
```diff
@@ -36,16 +36,19 @@ public static function extendPidListByChildren($pidList = '', $recursive = 0)
             return $pidList;
         }
 
+        /** @var \TYPO3\CMS\Core\Database\QueryGenerator $queryGenerator */
```
georgringer: please remove this line, those are not needed - use the phpstorm plugin `DynamicReturnTypePlugin`

georg-tiefenbrunn: Done. ThX for the plugin tipp :)

PR #274, opened 2017-03-03
Fixed wrong more-link language key in setup.txt

derhansen:
The default TypoScript setup contains the wrong language key for
 the more-link text


georgringer:
I changed the code as it makes no sense to override the german field actual

PR #281, opened 2017-03-17
Add Facebook Instant Articles feature to news

danielkestler:
Configures a new page type for Instant Articles RSS-Feed generation and supplies the content transformation and all the templates needed. See: https://github.com/medienreaktor/news_ia


georgringer:
sry don't plan to change there anything anymore

PR #287, opened 2017-03-27
Social ViewHelpers are missing use statements for GeneralUtility

fjacobi:
Added the missing ones.


fjacobi:
Already solved in #284 , #285 and #286

Issue #330, opened 2017-05-19
Missing typo3/class-alias-loader extra in composer.json in 4.3.0


dogawaf:
Hello

In news 4.3.0, the composer.json is missing an extra section for the autoloading of legacy class alias map.
This breaks TYPO3 installation in composer mode.

The composer.json should include this:
```
"extra": {
  "typo3/class-alias-loader": {
    "class-alias-maps": [
       "Migrations/Code/ClassAliasMap.php"
    ]
  }
},
```

As a workaround, adding this extra section in the global composer.json works.
In news 5, the legacy class alias map has been removed, so the issue is not existing.

I can make a pull request if it is OK.

Cheers.

georgringer:
yeah pls PR

2017-05-19 10:11 GMT+02:00 Rémy DANIEL <notifications@github.com>:

dogawaf:
I've created the pull request against the 4.3 branch.
Cheers.

georgringer:
sorry for failing to change that but I won't support the old branches anymore. therefore I am closing this issue.

PR #333, opened 2017-05-22
[BUGFIX] Fix composer autoloading for legacy classnames

dogawaf:
Resolves: #330


dogawaf:
Hello @georgringer 
Travis is not happy, I think it is because it runs tests with TYPO3 7 on the 4.3 branch of ext:news.
Is there something I can do?
Cheers.

PR #344, opened 2017-06-07
[BUGFIX] Prevent SQL error when filtering news by date

xperseguers:
This patch fixes a bug in the condition used to fetch the name of
the column which is used for datemenu.


xperseguers:
Unit tests are failing but the condition of the getter looks odd anyway, the `|| in_array` does not make sense

looks like you only want to allow values `datetime` and `archive` (would need `&&` instead of `||`) but in my case I expect `starttime` to be returned.

For me I'd suggest to remove the special check you have on the getter and adapt the unit test. I really don't get their purpose and this bugfix highlights the problem.

liayn:
I agree the condition should be `!empty() && in_array`

PR #345, opened 2017-06-07
Feature: Headline

marcobiedermann:
I've wrapped the news title in `f:format.raw` to output the headline.
This way you can add `&shy;` tags to the headline which allows you to set specific breakpoints.
Because headlines can get a bit long and are mostly bolder than regular text, they often tent to overflow / cut off.


georgringer:
Sorry but that will allow XSS which I really don't to have by default. Feel free to add this to your templates.

PR #368, opened 2017-07-04
[FEATURE] Merged constructor for proxy classes

Mabahe:


Mabahe:
Forward port again from the same modification for extender. Check for unit tests there https://github.com/evoWeb/extender/pull/6

Issue #394, opened 2017-08-04
SimplePrevNextViewHelper with custom sortField doesn't work


alsodenn:
If you set "sorting" or "title" as Sortfield then you get an error:

`Call to a member function getTimeStamp() on integer `

Because of line 191 of the Viewhelper:
`$whereClause .= sprintf(' AND %s %s %s', $sortField, $selector, $news->$getter()->getTimeStamp());`

A "title" or "sorting" has no TimeStamp...

georgringer:
do you want to create a PR for that?

ayacoo:
`            $whereClause .= sprintf(' AND %s %s "%s"', $sortField, $selector, $news->$getter());
            if (method_exists($news->$getter(), 'getTimestamp')) {
                $whereClause .= sprintf(' AND %s %s %s', $sortField, $selector, $news->$getter()->getTimestamp());
            }`

Is this maybe a solution?
- method_exists -> (PHP 4, PHP 5, PHP 7)
- String is default. If there is a integer, then set a integer

georgringer:
maybe just check if the object is from type datetime, then use the getTimeStamp, otherwise just use it as it is and yes security is a must

georgringer:
merged

PR #399, opened 2017-08-08
SimplePrevNextViewHelper with custom sortField doesn't work

ayacoo:
Relates #394


ayacoo:
First try with upstream syncing. If you need a cleaner version, please let me know. Then i create a new fork.

PR #411, opened 2017-08-27
Update ext_localconf.php

bobbel78:
Added DataHandler Hook for news records restrictions


georgringer:
can you give some more feedback to this change?

bobbel78:
I added a DataHandler Hook to restrict access to news with restricted categories, i.e. categories with special permissions for backend user groups. Before the change all users were able to edit all news even if categories were restricted.

Issue #420, opened 2017-09-08
Feature to disable/enable "Localization" field in the news list backend module.


arunspawoz:
Currently, there is no option to disable/enable the "Localization" field in the news list backend module.
Please see the screen below;

![news_modulelist 1](https://user-images.githubusercontent.com/20816321/30213267-f207bcce-94c5-11e7-93d9-b936e7ed10d7.png)

Please add a feature to manage this field.

Thank you.

georgringer:
fixed

PR #421, opened 2017-09-09
[FEATURE] Disable Localization in backend module via TSConfig (#420)

ayacoo:
Integration of a news TSConfig setting:
tx_news.module.localizationView = 0


ayacoo:
Added also an isset check for "filters." because of a local error message. Hopefully this is okay?

Classes/Controller/AdministrationController.php
```diff
@@ -289,7 +289,7 @@ public function indexAction()
         $dblist->script = GeneralUtility::getIndpEnv('REQUEST_URI');
         $dblist->thumbs = $this->getBackendUser()->uc['thumbnailsByDefault'];
         $dblist->allFields = 1;
-        $dblist->localizationView = 1;
+        $dblist->localizationView = $this->tsConfiguration['localizationView'] ? MathUtility::forceIntegerInRange($this->tsConfiguration['localizationView'], 0) : 1;
```
georgringer: just do a type cast to int, no need for usage of MathUtility

ayacoo: If i set (int) the default value for non-configured settings is 0, and that was a changed to the given behaviour. If you are agreed with this, i will change to integer.

PR #429, opened 2017-09-20
Update Index.rst

stefanisak:
So far I think `plugin.tx_news.feature.requireCHashArgumentForActionArguments  = 0` is the only way of getting this to work in TYPO3 >=8.5. If that's right, I thought it's worth mentioning...


georgringer:
plans are however to remove the no_cache and add a chash, see https://review.typo3.org/#/c/50264/

PR #430, opened 2017-09-20
[Bugfix] Grab translated Content Element id's

cptnk:
Just having getContentElementIdList returning the raw ID produces a problem when you lets say use dce plugin content with translations. Or different media elements on different languages. Returning the localized UID's fixes this issue.


georgringer:
As I don't know the side effects I added a new getter to get those ids

Issue #431, opened 2017-09-20
Icons should be registered in ext_tables.php file


morinfa:
When registered from ext_localconf.php file, the instantiation of IconRegistry class makes icons from other extensions fails to be displayed in the backend.
I guess this has to do with the handling of $GLOBALS['TCA'] in IconRegistry::initialize function.
When registered from ext_tables.php file, it is ok.

georgringer:
I am not sure that this is correct, IMO the bug you have is somewhere else, because icon registration must be done in ext_localconf.

morinfa:
If you look in workspaces module, it is done in ext_tables.php file.

georgringer:
changed with https://review.typo3.org/#/c/54020/

PR #437, opened 2017-09-28
Fixes issue #431

morinfa:
Register icons in ext_tables.php instead of ext_localconf.php in order to not screw icons from other extensions records in various views (e.g. list view, record references)


georgringer:
I close this PR as this is IMO the wrong approach, see comments in issue

PR #465, opened 2017-11-15
[BUGFIX] Correctly define file palettes for TYPO3 8

koehnlein:
TYPO3 8 does not support 'foreign_types' to define file fields. There is a new syntax 

```php
'overrideChildTca' => [
    'types' => [
        // ...
    ]
]
```

which seems not to be supported in TYPO3 7. So it's the best way to define the fields twice with both syntaxes to have the correct field in TYPO3 7 and TYPO3 8.


koehnlein:
Forget it - just found out, it's solved in TCA/Overrides

PR #477, opened 2017-11-30
[TASK] Allow tag localization

echavaillaz:


georgringer:
I will try to suport this with 7.1

echavaillaz:
You can close here, it is reolved with bdfde0b

Thanks.
Eric

PR #481, opened 2017-12-08
[BUGFIX] Make option list.paginate.prevNextHeaderTags work

christianfutterlieb:
This is quite a no-brainer. The condition that checks whether to generate the `<link rel="prev/next">` tags should not use `settings.list.paginate` but `settings.list.paginate.prevNextHeaderTags`.


georgringer:
Can you check styleci and fix that? Would be great

christianfutterlieb:
@georgringer I don't know why styleci mutters upon this change, the mentioned files are not touched by it? And I don't believe I should fix codestyle issues in this pr.. Can you give me a hint on how to proceed?

Issue #435, opened 2017-09-25
Logical problem with tags in newsRepository


derhansen:
In the newsDemand object, the property `tags` is defined as `ObjectStorage` - see [here](https://github.com/georgringer/news/blob/master/Classes/Domain/Model/Dto/NewsDemand.php#L40)

In the newsRepository, there is a [check](https://github.com/georgringer/news/blob/master/Classes/Domain/Repository/NewsRepository.php#L197) for `is_string` on the property `tags` of the demandObject and if so, it explodes the values of the `tags` property and searches by the uids of the given tags. If you set the tags as `ObjectStorage` in the demandObject, it is not possible to limit the query constraints be a given set of tags.

Either the `tags` property in the domandObject should be of type string or the newsRepository should also respect the scenario, that the `tags` property could be of type `ObjectStorage` and if so, add search constraints for all given tags.

I would prefer the last option. What is your opinion on that?

georgringer:
I would be a fan of changing it to string as that is the only non breaking change. can you do me a PR?

derhansen:
Yes, I will provide a PR for the change soon

georgringer:
merged

PR #483, opened 2017-12-10
Change datatype for tag in newsDemand to string

derhansen:
Refs: #435


georgringer:
Can you check the ci?

Issue #494, opened 2017-12-18
Tag list - no limit option


steffenk:
When displaying tag list, i can add tags, or display all.

I would like to have a limit set on CE level.

georgringer:
resolved

PR #497, opened 2017-12-28
#494 allow limit and offset for tags

steffenk:
Simply allow the fields in flexform to get limit + offset for tags


steffenk:
why does travis fail by a string-change patch???

PR #498, opened 2018-01-01
Fix/master/sitemap

### LINKED ISSUE #496: TxNewsSitemapGenerator should respect news type
Pages with the types "internal link" or "external link" will get the same paths as regular news. This is different from the behavior of LinkViewHelper which will `getInternalurl()` or `getExternalurl()` as a link parameter.

The proposed solution is as follows: In `getNewsItemUrl()` insert
```php
if ($newsRow['type'] == 1 && !empty($newsRow['internalurl'])) {
    $conf['additionalParams'] = $additionalParams;
    $conf['parameter'] = $newsRow['internalurl'];
}
if ($newsRow['type'] == 2 && !empty($newsRow['externalurl'])) {
    $conf['additionalParams'] = $additionalParams;
    $conf['parameter'] = $newsRow['externalurl'];
}
$link = htmlspecialchars($this->cObj->typoLink('', $conf));
```
The last line is from the original function right before the final `return` statement.

This has been successfully tested on TYPO3 8.7.8 with news 6.1.1 and internal news. We don't use external urls.

0bserver:
This fix resolves #496.

### DIFF
--- a/Classes/Hooks/TxNewsSitemapGenerator.php
+++ b/Classes/Hooks/TxNewsSitemapGenerator.php
@@ -184,11 +184,10 @@ class TxNewsSitemapGenerator extends AbstractSitemapGenerator
                 'returnLast' => 'url',
                 'useCacheHash' => true,
             ];
-            if ($newsRow['type'] == 1 && !empty($newsRow['internalurl'])) {
+            if (1 === (int)$newsRow['type'] && !empty($newsRow['internalurl'])) {
                 $conf['additionalParams'] = $additionalParams;
                 $conf['parameter'] = $newsRow['internalurl'];
-            }
-            if ($newsRow['type'] == 2 && !empty($newsRow['externalurl'])) {
+            } elseif (2 === (int)$newsRow['type'] && !empty($newsRow['externalurl'])) {
                 $conf['additionalParams'] = $additionalParams;
                 $conf['parameter'] = $newsRow['externalurl'];
             }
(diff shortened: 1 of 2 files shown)


Classes/Hooks/TxNewsSitemapGenerator.php
```diff
@@ -184,6 +184,14 @@ protected function getNewsItemUrl($newsRow, $forceSinglePid = null)
                 'returnLast' => 'url',
                 'useCacheHash' => true,
             ];
+            if ($newsRow['type'] == 1 && !empty($newsRow['internalurl'])) {
```
georgringer: please cast it to integer

Classes/Hooks/TxNewsSitemapGenerator.php
```diff
@@ -184,6 +184,14 @@ protected function getNewsItemUrl($newsRow, $forceSinglePid = null)
...
+                $conf['additionalParams'] = $additionalParams;
+                $conf['parameter'] = $newsRow['internalurl'];
+            }
+            if ($newsRow['type'] == 2 && !empty($newsRow['externalurl'])) {
```
georgringer: please use an elseif

Classes/Hooks/TxNewsSitemapGenerator.php
0bserver: Done …

Classes/Hooks/TxNewsSitemapGenerator.php
0bserver: … and done. Good point.

georgringer:
fixed with different commit

PR #519, opened 2018-01-17
[BUGFIX] Fix spelling

stucki:
Fix a small typo.

I found this change while searching for a similar solution in a different extension.

However, I've been wondering why the fix is applied to versions 9+ only? IMO it is also needed for every version 8.x.


georgringer:
8 does handle that internally as fallback

PR #546, opened 2018-01-25
Fixed calculation of numberOfPages

ddiesenreither:
Respect the offset when calculating the numberOfPages


georgringer:
I am reverting this issue because it is not that simple. e.g. when having limit of 1 and offset of 2, nothing is shown which is of course wrong.
if you still need this, please create a new issue and define your usecase, thx!

PR #547, opened 2018-01-26
Import command doesn't recognize the class to use

eliasfernandez:
On typo3 7.6.23 the argument passed as index is a numeric value but it is compared to a string.


georgringer:
the import command controller has been removed, therefore this PR is not needed anymore

Issue #433, opened 2017-09-22
Search Result without Pagination


valentinbuttkereit:
In SearchResult.html there is no pagination.

georgringer:
do you want to do a PR or sponsor that change?

ayacoo:
Hi @georgringer , i test a little bit for this implementation. I have searchform and searchresults one a single page.

I see that we can use the paginate viewhelper in SearchResult.html, too. It reacts completely to the given get params, maybe &tx_news_pi1[search][subject]=Test. So far so good

My only problem is: We must set the method to "get" in SearchForm.html and now i lost the pageUid and get the given referrer stuff in url. 

Do you have any ideas or hints for me?

georgringer:
you can readd the id but a real solution would be to change the implementation and remove all that referrer stuff there to simplify the GET params. but as said: this needs a sponsoring as I am not planning to do that

ayacoo:
Hi Georg, the implementation wasn't a problem for me and i would do this. But I just want implement a coordinated solution ;-)

Normally i would build a simple form (without fluid f:form) or build a copy form viewhelper without this rendering hidden component. But i prefer the first solution. Security shouldn't be a problem for this simple form.

georgringer:
that sounds ok, wanna start with a PR?

ayacoo:
I try to build a PR at the weekend :-)

georgringer:
thanks! I looked at it and gave a comment. therefore It seems that this can be done completly without any change to the extension and with template changes only. 

therefore I would propose to add this to the docs only and leave the default templates untouched. what do you say?

ayacoo:
For integrators might be okay. But what about the people who just install extensions and use the given features?

Or do you expect a certain level from your users? Then the extended documentation is absolutely sufficient :-)

robertvonhackwitz:
Hi @ayacoo could you explain how to implement pagination in search results?
I've tried following what you and Georg wrote but I think I haven't well understood :-(

jokumer:
related to #871

georgringer:
closed in favor of #871

PR #581, opened 2018-02-12
Extend searchresults with pagination

ayacoo:
Relates to https://github.com/georgringer/news/issues/433


Classes/Controller/NewsController.php
```diff
@@ -422,6 +422,7 @@ public function searchFormAction(
             'search' => $search,
             'overwriteDemand' => $overwriteDemand,
             'demand' => $demand,
+            'currentPageId' => $GLOBALS['TSFE']->id
```
georgringer: there is already `{pageData.uid}` which could be used

PR #608, opened 2018-03-16
Replace references of powermail in Readme

mkoitka:


mkoitka:
In the meantime already fixed by #610

PR #652, opened 2018-04-20
dont concatenate js/css files

Nemo64:
typo3 defaults to concatenate all css/js files. This is fine if the file is included on every page.
The `IncludeFileViewHelper` only adds css/js to specific pages (eg. where the news plugin is). This results into typo3 creating an entirely new css/js merge package with all other css/js files of the page added. This is a giant waste of bandwidth. If files really have to be added dynamically, then at least don't force the user to download all other files again also.

### DIFF
--- a/Classes/ViewHelpers/IncludeFileViewHelper.php
+++ b/Classes/ViewHelpers/IncludeFileViewHelper.php
@@ -56,20 +56,20 @@ class IncludeFileViewHelper extends \TYPO3\CMS\Fluid\Core\ViewHelper\AbstractVie
 
             // JS
             if (strtolower(substr($path, -3)) === '.js') {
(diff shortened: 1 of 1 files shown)


georgringer:
I won't merge that. for a proper integration you need custom css anyway and then I would just not include *any* CSS of ext:news

Nemo64:
I agree.

But the default css is there and you need to know that it isn't integrated like any other css file in typo3 AND that the default behavior will waste bandwidth and slow down your site and that you must change it for proper integration

I see the default behavior as a bug. ~ Although I'd argue that the default value of the addCss/Js functions is the bug.

Since other extensions like to wast user bandwidth as well without a way to disable that, this is currently the only correct solution in general: https://www.marco.zone/typo3-css-js-spam-verhindern

PR #662, opened 2018-05-03
[BUGFIX] selected in all views should be available in details view

sruegg:
Hello,

I think all images should be shown in detail view if the checkbox is checked "showinpreview" or the advanced selector is on "Show in all views".

Or am I missing something?

Thanks for all the work by the way :wink:!


wtfred:
@georgringer Sorry to bring this back, but i just notice this today, and i wanted to discuss something:
Before news 7.0.0, we had the checkbox "show in preview" on media
-> if checked, first media was used in list (in fluid `{newsItem.mediaPreviews.0}`)
-> if not, media was used in detail (in fluid `{newsItem.mediaNonPreviews}`)
But we could use in the detail template both `{newsItem.mediaPreviews}` and `{newsItem.mediaNonPreviews}`. That allowed us to display big images on top of the news with the mediaPreviews, and image gallery in bottom with mediaNonPreviews

Now, if we migrate to news 7+, `{newsItem.mediaNonPreviews}` contains both "preview" and "nonpreview" type of media.
I get it's consistant with the new list selector, but it would be nice to add in `Classes/Domain/Model/News.php` a method
```
/**
     * Get all media elements which are only for list views
     *
     * @return array
     */
    public function getMediaDetailOnly()
    {
        $configuration = [FileReference::VIEW_DETAIL_ONLY];
        return $this->getMediaItemsByConfiguration($configuration);
    }
``` 
so we could have the same behaviour as before by using `{newsItem.mediaDetailOnly}` instead of `{newsItem.mediaNonPreviews}`

I could make the PR if you find this relevant. Thanks for reading :)

PR #664, opened 2018-05-04
[BUGFIX] Forward from list to single when needed

helhum:
When list action is called, it means no action argument
for detail was set, which happens when using skipControllerAndAction
setting.

We then check for news argument and forward to detail action.

### DIFF
--- a/Classes/Controller/NewsController.php
+++ b/Classes/Controller/NewsController.php
@@ -194,6 +194,8 @@ class NewsController extends NewsBaseController
      */
     public function listAction(array $overwriteDemand = null)
     {
+        $this->forwardToDetailActionWhenRequested();
+
         $demand = $this->createDemandObjectFromSettings($this->settings);
         $demand->setActionAndClass(__METHOD__, __CLASS__);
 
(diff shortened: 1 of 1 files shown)


helhum:
## Scenarios **before** this change:

### 1. Page with a) plugin in detail view mode and b) plugin with list view mode

#### When I request this page with `tx_news[news]=1&tx_news[action]=Detail`

a) shows news with id 1
b) shows news with id 1

#### When I request this page with `tx_news[news]=1`

a) shows news with id 1
b) shows list of news


### 2. Page with a) plugin in detail view mode and b) plugin with list view mode (without detail)

#### When I request this page with `tx_news[news]=1&tx_news[action]=Detail`

a) shows news with id 1
b) shows list of news

#### When I request this page with `tx_news[news]=1`

a) shows news with id 1
b) shows list of news


## Same scenarios **after** this change:

### 1. Page with a) plugin in detail view mode and b) plugin with list view mode

#### When I request this page with `tx_news[news]=1&tx_news[action]=Detail`

a) shows news with id 1
b) shows news with id 1

#### When I request this page with `tx_news[news]=1`

a) shows news with id 1
**b) shows news with id 1**


### 2. Page with a) plugin in detail view mode and b) plugin with list view mode (without detail)

#### When I request this page with `tx_news[news]=1&tx_news[action]=Detail`

a) shows news with id 1
b) shows list of news

#### When I request this page with `tx_news[news]=1`

a) shows news with id 1
b) shows list of news

## Summary

There is indeed one change in behaviour and that is regarding the list plugin, when detail action is allowed but request does not have the action argument. For me this fixes a bug, since I don't expect the list action to show a list when the news argument is set. If I would want that, I would use the list mode without allowing detail action at all.

helhum:
Thanks!

PR #666, opened 2018-05-11
Improved compatibility for postgres installations

otisid:
As this is a minor correction, I don't know if it needs an specific issue, so feel free to reject/ask if issue needs to be open.
With these corrections, plugin will work in postgres installations (without it Typo3 raises exceptions and some user pages become impossible to be edited in Pages module and News administration module won't work).


georgringer:
thx for your work!

otisid:
Thank you (and other contributors) for the very useful plugin :)

PR #679, opened 2018-06-14
Update Detail.html

whiteroom:
Georg

Assuming I want to use HTML-Entities in dateFormat

```
plugin.tx_news._LOCAL_LANG {
        default {
			dateFormat = %-d.&#8202;%-m.&#8201;%Y
        }
}
```

Then we have to modify Detail.html and add <f:format.raw> in order to get wellformed html-output.
Otherwise the template spits out gibberish like `2.&#8202;1.&#8201;2017 `

Tested on 8.7.16 / latest tx_news
Best, Daniel


georgringer:
Do avoid side effects with f:format.raw I am not merging that pull requests. Change the templates if you need it, this is absolutly ok.

PR #690, opened 2018-06-23
Prevent empty link from LinkViewHelper

IchHabRecht:
If the link parameter is not accessible the returned url is empty.
Generating a link from an empty url results in a link to the current
page in Fluid TagBuilder api. This patch prevents the wrong link
to the current page and only returns the child rendering.


georgringer:
fixed already with https://github.com/georgringer/news/commit/b7d49dd95f8dbc58fdec452f31b57935e5769549

Issue #725, opened 2018-09-05
List.ical template: Output is not valid iCal format


hputzek:
## Bug Report

**Current Behavior**
The output from `Resources/Private/Templates/News/List.ical` is not valid. 

**Expected behavior/output**
Out should validate against https://icalendar.org/validator.html

**Environment**
- news version: 7.0.5
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: OSX

**Possible Solution**
I'll add a PR

hputzek:
Added a PR: https://github.com/georgringer/news/pull/726

georgringer:
solved

PR #726, opened 2018-09-05
[Bugfix] correct date formatting in iCal list template

hputzek:


hputzek:
## Bug Report

https://github.com/georgringer/news/issues/725

**Current Behavior**
The output from `Resources/Private/Templates/News/List.ical` is not valid. 

**Expected behavior/output**
Out should validate against https://icalendar.org/validator.html

**Environment**
- news version: 7.0.5
- OS: OSX

PR #737, opened 2018-09-16
Multilanguage: Get the ISO code information from page.conf

in-session:


georgringer:
I close this issue. when once 9 LTS is the minimum supported version I will switch to using the site handling code.

PR #748, opened 2018-10-03
fix dev dependencies for running tests with typo3 9.5

macjohnny:
fixes unit-tests running with typo3 9.5 and allows 9.5 compatibility


macjohnny:
see also https://github.com/bithost-gmbh/pdfviewhelpers/pull/91#issuecomment-426710930

georgringer:
fixed already, however tests still needs to be adopted

Issue #731, opened 2018-09-11
Query in SimplePrevNextViewHelper / fixed sys_language_uid of 0


jealouskid:
## Bug Report

Just a minor issue, but took me a while to find: The SimplePrevNextViewHelper is creating its queries using sys_language_uid set to 0 instead of $GLOBALS['TSFE']->sys_language_uid. This leads to random prev / next links in languages different to 0.

- TYPO3 version(s): 8.7.18
- news version: 7.0.5

**Possible Solution**
changing line 197 in SimplePrevNextViewHelper.php to
`$queryBuilder->expr()->eq('sys_language_uid', $queryBuilder->createNamedParameter($GLOBALS['TSFE']->sys_language_uid, \PDO::PARAM_INT)),`

weber-ebusiness:
In my opinion, this does not completely fix the bug.

There are two cases not covered:
1. Maybe some news records are not localized to the current language in `$GLOBALS['TSFE']->sys_language_uid`. In this case it should fallback to `sys_language_uid` = 0.
2. The news records are not localized at all, so all news records have `sys_language_uid` = -1.

georgringer:
if not fixed, not changing there stuff anymore

PR #768, opened 2018-10-18
[BUGFIX] Use language in SimplePrevNextViewHelper query

ayacoo:
Related #731


ayacoo:
Hint: In v9 we can use the Context API

https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/Context/Index.html
```
$context = GeneralUtility::makeInstance(Context::class);
$context->getPropertyFromAspect('language', 'id');
```

How would you want to implement that?

ayacoo:
So it would be better to even remove the line?

PR #792, opened 2018-11-07
[DOC] Add links to Settings.cfg

sypets:
- Remove commented out examples, because the links no longer exist
- Add link for issues and repo because that will then show up as
  "Related links" on docs.typo3.org
- Move sections to comply to generally used order

### DIFF
--- a/Documentation/Settings.cfg
+++ b/Documentation/Settings.cfg
@@ -1,12 +1,23 @@
+[general]
+project = News system
+release = 7.0.8
+version = 7.0
+copyright = 2018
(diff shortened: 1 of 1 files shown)


sypets:
See 
* https://github.com/xperseguers/t3ext-ig_ldap_sso_auth/blob/master/Documentation/Settings.cfg
* https://docs.typo3.org/typo3cms/extensions/ig_ldap_sso_auth/Index.html

The links for repo and issues, should now show up under "Related links" 
related

PR #800, opened 2018-11-17
[FEATURE] Add possibility to translate tags

### LINKED ISSUE #799: add possibility to translate tags
## Feature Request

One of our clients has many tags related to news. The use 2 languages and therefore I discovered, that the translation of tags is not possible (as opposed to categories). Are there certain reasons, why tags are non-translateable? Would it be easy to fix?

ewokhias:
I hope I did not forget anything. In simple tests, these changes seem to be sufficient.

Should close #799

### DIFF
--- a/Configuration/TCA/tx_news_domain_model_news.php
+++ b/Configuration/TCA/tx_news_domain_model_news.php
@@ -492,7 +492,8 @@ $tx_news_domain_model_news = [
                 'allowed' => 'tx_news_domain_model_tag',
                 'MM' => 'tx_news_domain_model_news_tag_mm',
                 'foreign_table' => 'tx_news_domain_model_tag',
-                'foreign_table_where' => 'ORDER BY tx_news_domain_model_tag.title',
+                'foreign_table_where' => ' AND (tx_news_domain_model_tag.sys_language_uid = 0 OR tx_news_domain_model_tag.l10n_parent = 0) ORDER BY tx_news_domain_model_tag.title',
+
                 'size' => 10,
                 'minitems' => 0,
                 'maxitems' => 99,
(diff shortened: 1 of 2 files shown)


Configuration/TCA/tx_news_domain_model_news.php
```diff
@@ -492,7 +492,8 @@
                 'allowed' => 'tx_news_domain_model_tag',
                 'MM' => 'tx_news_domain_model_news_tag_mm',
                 'foreign_table' => 'tx_news_domain_model_tag',
-                'foreign_table_where' => 'ORDER BY tx_news_domain_model_tag.title',
+                'foreign_table_where' => ' AND (tx_news_domain_model_tag.sys_language_uid = 0 OR tx_news_domain_model_tag.l10n_parent = 0) ORDER BY tx_news_domain_model_tag.title',
```
ayacoo: Would not it have to be this? tx_news_domain_model_tab.sys_language_uid IN (-1,0)

georgringer: I would go with the uid in(). wanna change that?

ewokhias: > Would not it have to be this? tx_news_domain_model_tag.sys_language_uid IN (-1,0) or better ###REC_FIELD_sys_language_uid### as placeholder?

I took the configuration from the categories field, but if the in()-approach is better, I'll fix it.

Is this correct?
`'foreign_table_where' => ' AND (tx_news_domain_model_tag.sys_language_uid IN (-1,0) OR tx_news_domain_model_tag.l10n_parent = 0) ORDER BY tx_news_domain_model_tag.title',`

PR #809, opened 2018-12-03
Fix link to page 1 in paginator when using native url rewriting in TYPO 9.5

wtfred:
Hello :)

People on Slack and stackoverflow are having troubles like "my code in config.yaml works for news details, it works for pagination too except for page 1"

This PR will normally fix it.
I only make changes in the "default" pagination template, maybe it will be required to make changes in other files if you are using another template for bootstrap style pagination.

PS: If you have advices to clean my git index before PR, i'll all ears, i'm not sure i'm doing it right.
I forked news some times ago, and keeping my fork up to date with [upstream](https://help.github.com/articles/configuring-a-remote-for-a-fork/) and [syncinc method](https://help.github.com/articles/syncing-a-fork/) But i'm not sure this is the best for git history


wtfred:
Careful, this PR will make the link to page 1 like this `/page-with-news-list/page/1`.
Maybe the expected behaviour is rather to link to `/page-with-news-list`, but if you don't pass parameters, that's when the link isn't rewrited...
So maybe the link to page 1 could link the "news list page uid", or maybe the problem with page 1 is more general and comes from a missing rewriting config, or a bug from the rewriting?

Issue #846, opened 2019-02-08
SlugService ignores duplicate titles (path_segment is not unique)


wazum:
## Bug Report

**Current Behavior**
Three news with the same title get the same value in path_segment

**Expected behavior/output**
The first news should get the path_segment transliterated from the title,
the second and third news with the same title should get a path_segment with a postfix like -1, -2
(the usual behaviour in TYPO3 at other places).

**Environment**
- TYPO3 version(s): 8.7
- news version: 7.0.8
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: Linux

**Possible Solution**
Check for duplicates in `\GeorgRinger\News\Service\SlugService::performUpdates`

**Additional context**
Upgrade from TYPO3 CMS 7 to 9

georgringer:
9 shouldn't use the slugservuce at all. wanna do a pr?

wazum:
Right, I updated the description, 'cause the problem actually occurs during an upgrade from 7 to 9 with TYPO3 CMS 8.
We have to solve this now, so I'll come back with a PR soon.

jokumer:
@wazum Nice solution. If you upgrade from earlier TYPO3 versions to TYPO3 9.5 with new introduced  routing, you may want to use same path segments as generated in realurl. Especially for news with similar titles, to ensure to have same url for preview version with realurl.

See my pullrequest https://github.com/georgringer/news/pull/849 which introduce migration of realurl alias to news slug/path_segment.

georgringer:
IMO this has been merged, so I close the issue

PR #847, opened 2019-02-08
[BUGFIX] Create a unique slug value in SlugService (#846)

wazum:


georgringer:
wanna take care of the conflicts? that would be great. thx

wazum:
I fixed the conflicts and used the `getUniqueValue` method in the `performRealurlAliasMigration` too, this makes at least sense for me (migrate as near as possible with existing RealURL values, but keep them unique too)

PR #872, opened 2019-03-19
FIX #870 Exception when importing tt_news with news_ttnewsimport

### LINKED ISSUE #870: Exception when importing tt_news with news_ttnewsimport
## Bug Report

**Current Behavior**
When importing a great number of tt_news-Entries (around 16.000) with news_ttnewsimport i got the following error message: 

Call to a member function getStorage() on null

and the import stops. 

**Expected behavior/output**
Importing all entries without exception

**Environment**
- TYPO3 version(s): [9.5.5]
- news version: [7.1.0]
- Is your TYPO3 installation set up with Composer (Composer Mode): [noi]
- OS: [OSX 10.11.6]
- news_ttnewsimport (Forge from jokumer)

**Possible Solution**
I found the reason and a working solution. 

One of the images in a tt_news-entry did have an error in the path: 
"uploads/pics/Bild:Leyendecker_9383.jpg"
Bild: is not a valid path and of course not a valid Storage. 

What triggers the exception: 

The NewsImportService-Class uses the $mediaItem['image'] array  in order to create a $file object: 
<pre>

try {
    $file = $this->getResourceFactory()->retrieveFileOrFolderObject($mediaItem['image']);
} catch (\TYPO3\CMS\Core\Resource\Exception\ResourceDoesNotExistException $exception) {
    $file = false;
}

</pre>
But the retrieveFileOrFolderObject- method of ResourceFactory class returns not only exceptions but in some cases NULL: 

<pre>

if (strpos($input, ':') > 0) {
    list($prefix) = explode(':', $input);
    if (MathUtility::canBeInterpretedAsInteger($prefix)) {
        // path or folder in a valid storageUID
        return $this->getObjectFromCombinedIdentifier($input);
    }
    if ($prefix === 'EXT') {
        $input = GeneralUtility::getFileAbsFileName($input);
        if (empty($input)) {
            return null;
        }

        $input = PathUtility::getRelativePath(Environment::getPublicPath() . '/', PathUtility::dirname($input)) . PathUtility::basename($input);
        return $this->getFileObjectFromCombinedIdentifier($input);
    }
    return null;
}

</pre>

so when a path contains a colon the result is not an exception but NULL. This triggers the exception which terminates the import. 

--

A working solution could be to change the following lines in NewsImportService (line 196)

<pre>
// no file found skip processing of this item
if ($file === false) {
    continue;
}
</pre>

to 

<pre>
// no file found skip processing of this item
if ($file === false || $file === NULL) {
    continue;
}
</pre>

The solution works for me, i could import all 16.000 tt_news entries. 

**Additional context**

lisardo2:
As written in the issue #870

### DIFF
--- a/Classes/Domain/Service/NewsImportService.php
+++ b/Classes/Domain/Service/NewsImportService.php
@@ -203,7 +203,7 @@ class NewsImportService extends AbstractImportService
                 }
 
                 // no file found skip processing of this item
-                if ($file === false) {
+                if ($file === false || $file === NULL) {
                     continue;
                 }
 
@@ -250,7 +250,7 @@ class NewsImportService extends AbstractImportService
                 }
 
                 // no file found skip processing of this item
-                if ($file === false) {
+                if ($file === false || $file === NULL) {
                     continue;
                 }
 
--- a/Classes/Domain/Service/NewsImportService.php
+++ b/Classes/Domain/Service/NewsImportService.php
@@ -199,11 +199,11 @@ class NewsImportService extends AbstractImportService
                 try {
                     $file = $this->getResourceFactory()->retrieveFileOrFolderObject($mediaItem['image']);
                 } catch (\TYPO3\CMS\Core\Resource\Exception\ResourceDoesNotExistException $exception) {
-                    $file = false;
+                    $file = null;
                 }
 
                 // no file found skip processing of this item
-                if ($file === false || $file === NULL) {
+                if ($file === null) {
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


Classes/Domain/Service/NewsImportService.php
```diff
@@ -203,7 +203,7 @@ protected function hydrateNewsRecord(
                 }
 
                 // no file found skip processing of this item
-                if ($file === false) {
+                if ($file === false || $file === NULL) {
```
DavidBruchmann: Isn't it better to set $file in line 202 to `null` instead of `false`?
Then the condition could be only `if ($file === null) {` (null lowercase) and cover all negative cases, or do I mistake about it?

lisardo2: Should work i presume ...

lisardo2: Shorter Code. I will change it.

lisardo2:
tested my import once again with the fix. Worked, all 16.000 imported.

Issue #907, opened 2019-05-09
It's not possible to use option orderByAllowed


DanielSiepmann:
## Feature Request

We would like to create new custom plugins for news within our installation, e.g. `news_recent`. This is a use case provided by TYPO3. Poorly the NewsController has hardcoded values for `news_pi`, preventing configuration of Option `orderByAllowed`. This is prevented to be overwritten by Flexforms. As the original value for reset is always fetched from `pi1`, it's not possible to define a different one for another plugin signature.

We therefore would propose to follow Extbase convention, and use `$this->extensionName` and `$this->pluginName` instead of hardcoded values.

codegm:
Please revert!

In typo3 v9 "pluginName" is always null, that result in tssettings is missing overrides from typoscript.

In v10 "extensionName" was removed, so would be breaking anyway in the next months. 
https://github.com/TYPO3/TYPO3.CMS/commit/a8b673daf995b024580d393f5ebada6ea82bc6b1

DanielSiepmann:
So instead of reverting, I would suggest to replace it with a call to the request to fetch the information.

PR #927, opened 2019-06-16
Use stable API to fetch extension name and plugin name

DanielSiepmann:
The existing approach was removed with
https://forge.typo3.org/issues/87627.

Relates: #907, #908


DanielSiepmann:
I didn't run any tests, but installed TYPO3 v8 and used auto completion.

codegm:
The extbase confManager expects 'Pi1' when i get this right from the code.

'news_pi1' was then wrong…!? So my typoscript config was not expected behavior since years

georgringer:
IMO I had to revert this once merged and as the PR is quite old I won't merge it

PR #934, opened 2019-07-01
[BUGFIX] Remove "default" richtextConfiguration from TCA

ikonplant:
Reasons:
1. typo3/sysext/core/Classes/Configuration/Richtext.php, public function getConfiguration uses "default" anyway at the very last point of the fallback-chain if no richtextConfiguration was explicit set via e.g. PageTS or TCA
2. function getConfiguration was broken till now and did not take in account the settings defined via TCA
3. this patch: https://review.typo3.org/c/Packages/TYPO3.CMS/+/61193/3 (under review) will fix the issue of ignoring TCA settings; but the value of RTE.default.preset, set within custom extension via PageTS, would be ignored, since "default" is set in the TCA of the news-extension. 

Get more details at the commit message within provided link & feel free to ask

### DIFF
--- a/Configuration/TCA/Overrides/sys_category.php
+++ b/Configuration/TCA/Overrides/sys_category.php
@@ -135,7 +135,6 @@ $newSysCategoryColumns = [
         'config' => [
             'type' => 'text',
             'enableRichtext' => true,
-            'richtextConfiguration' => 'default',
         ],
     ],
 ];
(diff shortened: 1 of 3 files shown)


georgringer:
@josefglatz please an opinion on that

ikonplant:
Could you please (re)check that?
The mentioned patch in TYPO3 core is merged and will be released with v9.5.12

ikonplant:
Sorry, the last information is wrong.
The patch hasn't been cherry-picked for 9.5.
But still we can remove "default" without any impact (my first comment, reason 1)

PR #947, opened 2019-07-29
[Bugfix] Constructor in subsequent proxyclasses

Crontis:
The constructorLines property is never unsetted in ClassCacheManager.
This results in added constructors for subsequent classes if this
property is set.

Resetting this property to an empty array fixes this issue.


froemken:
Was already solved in EXT:extender
https://github.com/evoWeb/extender/commit/23c6c076d2d1f86c4aeac43a4ee92fe6d275e0aa

PR #1021, opened 2019-08-23
Put f:widget.uri arguments in quotation marks to prevent exception

ddiesenreither:
After clearing cache, this caused an exception because of wrong arguments.


georgringer:
widgets have been removed, closing this pr

PR #1052, opened 2019-09-17
[FEATURE] Add Google News mode to xml sitemap provider

liayn:


liayn:
TODO: add google news sitemap template file to docs as example

Issue #1061, opened 2019-10-09
Exception on Backend when trying to edit a category.


lgescobar:
## Bug Report

**Current Behavior**
I am not able to edit a `sys_category` record, because an exception is thrown in backend.

**Expected behavior/output**
I expect to be able to edit that record 

**Environment**
- TYPO3 version: **9.5.9**
- PHP version: **7.3.10** and **7.3.9**
- news version: **7.3.0** (probably affected since 7.1.0)
- Is your TYPO3 installation set up with Composer (Composer Mode): **yes**
- OS: macos 10.14.6, Docker Desktop for Mac 2.1.0.3 (38240) + ddev 1.11.1 (homebrew)

**Possible Solution**
Pull request #1062 

**Additional context**
After some analysis I could find the error. In TYPO3 9 there is the new field `slug` (since EXT:news 7.1.0) for types `sys_category` and `tx_news_domain_model_tag` and its TCA configuration contains a display condition (`displayCond` key in column configuration). 

The problem with it is, that the function `\TYPO3\CMS\Core\Compatibility\PseudoSiteTcaDisplayCondition::isInPseudoSite()` is intended to work with "pages" records and when editing a category this function checks if the page with the same `uid` as the category record has a pseudo site object or a site object attached. Most of the time this causes no errors because a page with the same `uid` exists and is attached to a PseudoSite object. (Statistically the number of category records is less than the pages' one and that pages live throughout the whole system's lifetime)

In my case I tried to edit the record `sys_category:84` and there was the records `pages:6` (`pid` field set to `1`) and `pages:84` (`pid` field set to `6`) both with `deleted` and `hidden` fields set to `1` as well as `pages:1` which is an active "siteroot" page placed on the root level.

snk-typo3-team:
This also happens if you click the "NEW" button inside the record. 

`$defaultLanguagePageId` which the PseudoSiteFinder uses is 0 because there is a typecast on `$parameters['record']['uid']` which in this case is not the uid but the `NEW5dce948fd0d77252477182` string.

georgringer:
resolved since couple versions

PR #1062, opened 2019-10-09
[BUGFIX] Check for Site or PseudoSite object of containing page

lgescobar:
Bug fixed by using a kind of a composition. The new class extracts the page record of the page containing the given record by using its `pid` field and calls the original display condition with updated parameters.

To avoid breaking things up, the new class just returns the proper value it the internal core class cannot be instantiated (should be gone with TYPO3 v10).

Related to issue #1061


matis (approved):
Testet on TYPO 3 9.5.21 with EXT News  8.4.0: fixed the problem

georgringer:
the displayCond has completly changed, therefore closing this PR

PR #1075, opened 2019-10-17
*fixed: bootstrap 4 responsive images (https://getbootstrap.com/docs/…

dia-pko:
*fixed: bootstrap 4 responsive images (https://getbootstrap.com/docs/4.3/content/images/#responsive-images)


Resources/Private/Templates/Administration/Donate.html
```diff
@@ -112,13 +112,13 @@ <h3>{f:translate(key:'LLL:EXT:news/Resources/Private/Language/locallang_be.xlf:a
...
-                 class="img-responsive"/>
```
georgringer: there is no bootstrap 4 in the backend?!

cboltz:
This change is needed in combination with the latest `bootstrap_package` (11.0) which indeed includes bootstrap 4.

https://getbootstrap.com/docs/4.0/migration/#images confirms that `.img-responsive` was renamed to `.img-fluid`.

If you want to make this a backwards-compatible change, one option would be to use `class="img-responsive img-fluid"`

georgringer:
as templates can easily changed by integrators and the change would lead to possible regressions, I am closing it

PR #1076, opened 2019-10-18
Added "Extend Flexform" Example Code for 9x

TheNaderio:
PATH_site is deprecated and will be removed in v10.
Added `Environment::getPublicPath()`and a slash before `/typo3conf/...`


georgringer:
Please make the pr against branch 8.0 as this is 9 and 10 only, that would be great!

TheNaderio:
Moving into other branch branch

PR #1089, opened 2019-10-23
[BUGFIX] Use SlugHelper::generate in order to considerate slug field TCA configuration

### LINKED ISSUE #1088: Slug upgrade wizard does not consider slashes
## Bug Report

**Current Behavior**
When using the upgrade wizard `Updates slug field "path_segment" of EXT:news records` slashes within a news record title are not removed. 

Example: 
Title: "Some nice news record / article with a slash"
Slug: "some-nice-news-record-/-article-with-a-slash"

**Expected behavior/output**
Slug: "some-nice-news-record-article-with-a-slash"
Slashes need to be removed, otherwise the routing will crash with something like:

```
(1/1) Symfony\Component\Routing\Exception\InvalidParameterException

Parameter "tx_news_pi1__news" for route "tx_news_pi1_0" must match "[^/]++" ("projektfilm-stellt-arbeit-und-leitmotive-des-dortmunder-profil-fuer-inklusionsorientierte-lehrer/-innenbildung-doprofil-vor" given) to generate a corresponding URL.
```

**Environment**
- TYPO3 version(s): 9.5.10
- news version: 7.3.1
- OS: Ubuntu 16

**Possible Solution**
Using SlugHelper::generate() instead of SlugHelper::sanitize() would take the TCA config `replacements` in consideration (related to #832). A simple str_replace would do the job, too.

**Additional context**
Here's a simple SQL query which fixes the issue quick and dirty:

`UPDATE tx_news_domain_model_news SET path_segment = replace(path_segment, '/', '');`

fnagel:
When a news title contains a slash the slug update wizard will not
remove those. This will cause broken routing. With this change the TCA
configuration is taken into consideration which results in removal of
slashes.

Fixes #1088

### DIFF
--- a/Classes/Service/SlugService.php
+++ b/Classes/Service/SlugService.php
@@ -33,15 +33,6 @@ class SlugService
     }
 
     /**
-     * @param string $string
-     * @return string
-     */
-    public function generateSlug(string $string): string
-    {
-        return $this->slugService->sanitize($string);
-    }
-
-    /**
      * @return int
      */
     public function countOfSlugUpdates(): int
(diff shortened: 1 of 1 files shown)


pulponair:
@georgringer any chance to incorparete this to 7.3.2?

Issue #1100, opened 2019-11-05
Typo3 10 Compatibility


rfc2822:
Hi,

I have made some work for Typo3 10 compatibility in my fork: https://github.com/rfc2822/news (maybe I'll add another commits)

However, it would need some testing and I don't know whether it's still compatible with older versions because I only use it for a Typo3 10 project.

Is there a way how to help contribution Typo3 10 support? Shall I send a PR when I'm finished? But it would need more testing.

georgringer:
Pls checkout https://github.com/georgringer/news/tree/8.0?files=1

rfc2822:
Thanks, I'll test the 8.0 branch which seems to work with Typo3 10. Is there still any specific work to do?

georgringer:
checkout https://github.com/georgringer/news/projects ;)

PR #1101, opened 2019-11-08
[WIP] Typo3 10 compatibility

rfc2822:
**Work for Typo3 10 compatibility, should go to new 10.x branch.**

Maybe not compatible with older versions anymore. (Would have to be tested)

See #1100


georgringer:
closing the PR as fixed

Issue #1235, opened 2020-04-28
DataMapper QueryInterface injection workaround should be removed


bnf:
DataMapper isn't really supported to be injected (it is internal code), still the issue is fixed in TYPO3.CMS 10.4.1 due to the following change:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/64319

Therefore the following code can be removed once the dependency requirement is upgraded to `^10.4.1`:
https://github.com/georgringer/news/blob/2df2e7463b14f1da7c9ebc0364dea221064c1d77/Configuration/Services.yaml#L33-L35

bnf:
Hm… I guess this is hard to change as we can't exclude 10.4.0 in ext_emconf (for classic mode installations) without removing support for v9.

If we add a `conflicts` entry like the following:
```
'depends' => [
    'typo3' => '9.5.9-10.9.99',
],
'conflicts' => [
    'typo3' => '10.0.0-10.4.0',
],
```
…then the ExtensionManager is buggy and reports the opposite:

```
Your TYPO3 version is higher than this extension requires.
It requires TYPO3 versions 10.0.0 - 10.4.0
```

georgringer:
removed with e82cff29b730c760abfad57227eb28556b481694

PR #1236, opened 2020-04-28
[WIP][TASK] Drop DataMapper workaround for QueryInterface injection

bnf:
This has been fixed in TYPO3.CMS in version 10.4.1:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/64319

TODO:
 * How to exclude 10.4.0 for classic mode?
   excludes in ext_emconf doesn't work

Resolves: #1235


georgringer:
@bnf conflict won't work. So I would suggest just waiting another 10.4.x release and merge it then anyway

georgringer:
fixed it, thx for your PR

PR #1273, opened 2020-05-11
[BUGFIX] Use correct module routes

### LINKED ISSUE #1271: Backend Modules not loading
## Bug Report

**Current Behavior**
When using the search form within the news administration module an Exception is thrown.
```
#1425389240 TYPO3\CMS\Backend\Routing\Exception\ResourceNotFoundException
The requested resource "/module/web/NewsAdministration" was not found.
in /var/www/html/public/typo3/sysext/backend/Classes/Routing/Router.php line 78
```

**Expected behavior/output**
The results for the search should be loaded without an exception.

**Environment**
- TYPO3 version: 9.5.16
- news version: 8.2.0
- Composer Mode: yes
- OS: Linux (drud/ddev-webserver:v1.13.0)

**Possible Solution**
The module route that gets generated by TYPO3 for the main news module is just `/web/NewsAdministration`. When manually changing the hard coded usages of the route from `/module/web/NewsAdministration` to `/web/NewsAdministration` the module loads and the search results are shown.

**Additional context**
We only recently upgraded to News 8.0.2 and were using 7.5.3 before. The version jump was done with the recent changes in TYPO3 9.5.16 (multi domain setup with news detail pages on different sites).

Here is an exerpt of the `var_dump` done in the `TYPO3\CMS\Backend\Routing\Router->match` function, showing the config for the target module:
```
["web_NewsAdministration"]=>
  object(TYPO3\CMS\Backend\Routing\Route)#1245 (2) {
    ["path":protected]=>
    string(24) "/web/NewsAdministration/"
    ["options":protected]=>
    array(4) {
      ["module"]=>
      bool(true)
      ["moduleName"]=>
      string(22) "web_NewsAdministration"
      ["access"]=>
      string(10) "user,group"
      ["target"]=>
      string(54) "TYPO3\CMS\Extbase\Core\Bootstrap::handleBackendRequest"
    }
  }
```
This lead to the idea to change hard coded routes like this one in `Resources/Private/Templates/Administration/Index.html::49`
```
<input type="hidden" name="route" value="/module/web/NewsAdministration">
```

jonakieling:
Hi there,

this fixes #1271
Some hard coded backend module routes had the /module prefix which does not work in TYPO3 9.5.16 and presumably any version of 9.5. At least I could not find any changes related to backend module routing.

This fixes our TYPO3 9.5.16 instance where the previous routes resulted in an exception being thrown. See #1271 for details.

I'm curious where the /modules comes from or where it would be needed. Maybe I'm just missing something obvious.

Kind regards
Jona

### DIFF
--- a/Classes/Backend/RecordList/RecordListConstraint.php
+++ b/Classes/Backend/RecordList/RecordListConstraint.php
@@ -31,7 +31,7 @@ class RecordListConstraint
     public function isInAdministrationModule(): bool
     {
         $vars = GeneralUtility::_GET('route');
-        return $vars === '/module/web/NewsAdministration';
+        return strpos('/module/web/NewsAdministration', $vars) !== false;
     }
 
     public function extendQuery(array &$parameters, array $arguments)
--- a/Resources/Private/Templates/Administration/Index.html
+++ b/Resources/Private/Templates/Administration/Index.html
@@ -46,7 +46,7 @@
 	<f:if condition="{showSearchForm}">
 		<f:form id="administrationForm" name="demand" object="{demand}" method="get" data="{autoSubmitForm:autoSubmitForm}">
 			<input type="hidden" name="formSubmitted" value="1">
-            <input type="hidden" name="route" value="/module/web/NewsAdministration">
+            <input type="hidden" name="route" value="/module/web/NewsAdministration/">
             <input type="hidden" name="token" value="{moduleToken}">
 			<input type="hidden" name="id" value="{page}">
 			<f:form.checkbox style="display:none" property="selectedCategories" value="0" />
--- a/Classes/Controller/AdministrationController.php
+++ b/Classes/Controller/AdministrationController.php
@@ -546,7 +546,7 @@ class AdministrationController extends NewsController
         /** @var \TYPO3\CMS\Backend\Routing\UriBuilder $uriBuilder */
         $uriBuilder = GeneralUtility::makeInstance(\TYPO3\CMS\Backend\Routing\UriBuilder::class);
 
-        $returnUrl = $uriBuilder->buildUriFromRoutePath('/web/NewsAdministration/', [
+        $returnUrl = $uriBuilder->buildUriFromRoutePath('/module/web/NewsAdministration/', [
             'id' => $this->pageUid,
             'token' => $this->getToken(true)
         ]);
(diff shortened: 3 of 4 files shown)


georgringer:
@liayn wanna test that in 10.x?

liayn:
The search worked flawlessly for us in the latest 8.x versions. Will try it again with this change applied.

liayn:
This change breaks in v10:

The requested resource "/web/NewsAdministration" was not found.

jonakieling:
Added a version check for getting the module path and adjusted two occurrences where the hard coded path was used.

I've checked the search form in 9.5.17 and 10.4.2. Looks good so far.

Classes/Backend/RecordList/RecordListConstraint.php
```diff
@@ -31,6 +32,11 @@ class RecordListConstraint
     public function isInAdministrationModule(): bool
     {
         $vars = GeneralUtility::_GET('route');
+        $version = (new Typo3Version())->getMajorVersion();
+        if ($version < 10) {
+            return $vars === '/web/NewsAdministration/';
+        }
+
```
liayn, line 39: what about using str(r)pos() instead of a version check for this? would be much simpler here.

`return strpos('/web/NewsAdministration/', $vars) !== false;`

liayn:
Btw: the related core change is https://review.typo3.org/c/Packages/TYPO3.CMS/+/60433

liayn:
A different solution would also be to set the desired path manually, as outlined in the deprecation rst file.

Resources/Private/Templates/Administration/Index.html
```diff
@@ -46,7 +46,7 @@ <h1>
 	<f:if condition="{showSearchForm}">
 		<f:form id="administrationForm" name="demand" object="{demand}" method="get" data="{autoSubmitForm:autoSubmitForm}">
 			<input type="hidden" name="formSubmitted" value="1">
-            <input type="hidden" name="route" value="/module/web/NewsAdministration">
+            <input type="hidden" name="route" value="/module/web/NewsAdministration/">
```
liayn, line 49: the slash is too much

jonakieling: Yep, just tested it with 9 LTS and 10 LTS. It is needed with 9 and breaks with 10.

jonakieling: Looks like the version check is unfortunately needed after all.

jonakieling: I'll move my fork to a different branch to avoid the commit spam and report back when everything is tested and working with both 9 and 10.

jonakieling: The slash is assumed by the older version in https://review.typo3.org/c/Packages/TYPO3.CMS/+/60433/5/typo3/sysext/core/Classes/Utility/ExtensionManagementUtility.php#b816. I adjusted the manual module path to include the slash. After the correct TCA was loaded without cache it now works with 9 LTS and 10 LTS.

liayn:
Looks good by reading

jonakieling:
Hi @liayn 

Thanks for your support in this! After checking this with flushed TCA caches the most recent commits work with both TYPO3 9 LTS and 10 LTS.

jonakieling:
Latest update fixes the action buttons for creating and editing records on the upper button bar.

PR #1312, opened 2020-07-09
[WIP][FEATURE] Add cached plugin

CDRO:


CDRO:
This change solves nothing, the problem we encountered was due to a misconfiguration on our side.

PR #1318, opened 2020-07-17
[DOCS] Fix code blocks in TypoScript settings

sypets:
- Fix the indenting. For some codeblocks, they had mixture of
  tabs and spaces or were not sufficiently indented
  which resulted in indenting levels being rendered incorrectly
  (e.g. list.rss, list.media)
- 3 spaces indenting is now used for all TS codeblocks. This
  conforms to the convention for TYPO3 docs projects but not
  to CGL for TypoScript (which is 2 spaces). Though it may be
  desirable to have 2-space indenting within codeblocks or
  TABs, this makes editing more difficult and may result in
  accidentally rendering errors.
- Add highlight directive for TypoScript page. (As configured
  in Includes.txt, PHP is the default language. The highlight
  directive sets default language and should help with correct
  syntax highlighting).
- Code samples with inline code are converted to code blocks
  to ensure that all examples look similar.


georgringer:
@sypets can you take care about the conflict?

sypets:
ok. Resolved conflicts and fixed errors introduced with resolving conflicts ;-)

PR #1326, opened 2020-08-19
[BUGFIX] Add missing default value for sitemap columns

echavaillaz:
When saving a news in the TYPO3 backend, there is no default value for the sitemap columns.
So the priority is set to 0.0 instead of 0.5.


georgringer:
hey, solved that, thanks for your PR

Issue #1336, opened 2020-09-11
Documentation: Index news records, don't index page itself (SEO)


sebkln:
By default, the page which contains the detail view plugin will be listed in the SEO sitemap.
But in most cases, you don't want the page itself to be indexed.

If you just disable *Index this page* in the page properties, the robots meta tag with *noindex* value will also be applied to the news records.

I will add a short section to the existing SEO documentation. The included TypoScript condition will use the MetaTag API of TYPO3 to replace the meta tag for news records.

PR #1337, opened 2020-09-11
[DOC] Index news records, don't index page itself (SEO)

sebkln:
Resolves: #1336

### DIFF
--- a/Documentation/AdministratorManual/BestPractice/Seo/Index.rst
+++ b/Documentation/AdministratorManual/BestPractice/Seo/Index.rst
@@ -241,3 +241,21 @@ If you are building a language menu and want to check if the news record is avai
            </f:if>
        </f:for>
    </ul>
+
+Robots: allow indexing of news records, but set 'noindex' to the detail page
+----------------------------------------------------------------------------
+By default, the detail page will be listed in the SEO sitemap.
+But in most cases, you don't want the page itself to be indexed by search engines (means: without a news record to be shown by the plugin).
+
+If you just disable *Index this page* (`no_index`) in the page properties, the robots meta tag with *noindex* value will also be applied to the news records.
+
+Solution: You can use the following TypoScript condition to allow search engines to index the page again, if a news record is rendered:
+
+.. code-block:: typoscript
+
+   [request.getQueryParams() ['tx_news_pi1'] > 0]
+       page.meta.robots = index,follow
+       page.meta.robots.replace = 1
+   [global]
+
+An important part is the `replace` option. The MetaTag API of TYPO3 will then replace tags which were set before.


georgringer (approved):

DanielSiepmann:
In my opinion that solution is conceptional wrong.
A clean website should never link to the site without any news entry.
I guess the only issue is that this site shows up in a sitemap.

The real solution should be to exclude the page from sitemap. That way it would not occur anywhere and should not be detected for indexing. That would prevent any hacks.
All news should be provided in a separate sitemap.

DanielSiepmann:
E.g. via:
```
plugin.tx_seo {
    config {
        xmlSitemap {
            sitemaps {
                pages {
                    config {
                        additionalWhere := appendString( AND uid != 4)
                    }
                }
            }
        }
    }
}
```

Or an additional field in pages "exclude_from_sitemap" for example.

DavidBruchmann:
> The real solution should be to exclude the page from sitemap

I understand what you mean @DanielSiepmann as it's related to avoid wrong links.  

Just as consideration I want to mention a variation of menus and thinkable for sitemaps too:  
For news it might be desired having news-articles listed directly inside and not as separated block / section / page.  
Certainly that's applicable for some other extensions too.

sebkln:
I agree @DanielSiepmann.

Yes, the original issue is the news detail page in the sitemap.

Disabling *Index this page* seems to be the obvious and common solution, at least for people I talked to. This will then result in the new issue, which I tried to fix here.

Your solution is of course cleaner, excluding the page directly in the sitemap configuration.

Yes, conditions should be avoided if possible. Your `additionalWhere` is the better approach in this case.
The only (small) benefit of the condition would be the general application to all subsequently created detail pages (no need for the integrator to add each UID to the config).

As this makes my suggested approach not a good practice, we should not add it to the news documentation. Do you want to add your solution instead?

haassie:
Don't forget instances that have list and detail content element on the same page maybe that is worth mentioning.

IMHO: If you have a separate detail page, a 404 should be returned when you access the detailpage without a proper news-item as there is no page without the article. With a 404 you don't have an issue with robots as well. Only thing you have to take care of is the sitemap indeed and I do like the suggestion of @DanielSiepmann in that case.

josefglatz:
> Don't forget instances that have list and detail content element on the same page maybe that is worth mentioning.
> 
> IMHO: If you have a separate detail page, a 404 should be returned when you access the detailpage without a proper news-item as there is no page without the article. With a 404 you don't have an issue with robots as well. Only thing you have to take care of is the sitemap indeed and I do like the suggestion of @DanielSiepmann in that case.

I have to say that Google Analytics annoys you when you add the news detail page to sitemap.xml but throwing 404 when no news record is available for display – if you open "blank" detail page. So that's problematic, since the SEO-person annoys you

josefglatz:
## Just for the records: I actualy do it like so:

**Fluid** (News Detail.html template):

```
<theme:metaTag.index />
```

**ViewHelper** \Supseven\Theme\ViewHelpers\MetaTag\IndexViewHelper:

```
<?php
declare(strict_types = 1);

namespace Supseven\Theme\ViewHelpers\MetaTag;

use TYPO3\CMS\Core\MetaTag\MetaTagManagerRegistry;
use TYPO3\CMS\Core\Utility\GeneralUtility;
use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
use TYPO3Fluid\Fluid\Core\ViewHelper\AbstractViewHelper;
use TYPO3Fluid\Fluid\Core\ViewHelper\Traits\CompileWithRenderStatic;

/**
 * Class IndexViewHelper
 *
 * Use this viewhelper to override robots with "index,follow".
 *
 * Useful if you set noindex on the detail page for e.g. ext:news in page settings but you want to override the value of
 * it within your extbase plugin's detailView.
 */
class IndexViewHelper extends AbstractViewHelper
{
    use CompileWithRenderStatic;

    public static function renderStatic(
        array $arguments,
        \Closure $renderChildrenClosure,
        RenderingContextInterface $renderingContext
    ) {
        $metaTagManager = GeneralUtility::makeInstance(MetaTagManagerRegistry::class)->getManagerForProperty('robots');
        $metaTagManager->removeAllProperties();
        $metaTagManager->addProperty('robots', 'index,follow', [], true);
    }
}

```

dev-rke:
I used the approach mentioned in the docs some weeks ago, disabled indexing on the page and re-enabled it using TypoScript when GET parameter tx_news_pi1 is set.
This leads to the special effect, that my canonical tags won't work anymore, due to this line of the core seo extension:
https://github.com/TYPO3/TYPO3.CMS/blob/v10.4.9/typo3/sysext/seo/Classes/Canonical/CanonicalGenerator.php#L61
The SEO extension doesn't care about TypoScript and just checks the configuration of the page, so it disables canonical tags, when page indexing is disabled.
As TYPO3 still allows usage of trailing slashes on urls this leads easily to duplicate content warnings. :-/

My next approach will be to remove the page from sitemap, as it sounds more promising to solve the source of the issue than implementing more workarounds. If it won't work as expected, my last approach will be to ignore that SEO shit. ;-)

georgringer (approved):

Issue #1305, opened 2020-06-26
Wrongly escape opengraph og:url property on non speaking url


fiotakis:
## Bug Report

**Current Behavior**

We have a site with the traditional urls (of the form ?id=123&....).

The opengraph og:url property has a wrongly escaped content value. The ampersands are escaped, which generates wrong urls.

Correct URL: 
https://www.cadlab.tuc.gr/index.php?id=12493&tx_news_pi1%5Bnews%5D=19491&tx_news_pi1%5Bcontroller%5D=News&tx_news_pi1%5Baction%5D=detail&cHash=a61639f720d6b1bc0fb2e46d619276b5

URL generated for the og:url property:
https://www.cadlab.tuc.gr/index.php?id=12493&amp;amp;tx_news_pi1%5Bnews%5D=19491&amp;amp;tx_news_pi1%5Bcontroller%5D=News&amp;amp;tx_news_pi1%5Baction%5D=detail&amp;amp;cHash=a61639f720d6b1bc0fb2e46d619276b5

Notice the ampersands which are generated as `&amp;amp;`

This results in all sorts of problems when sharing a news detail page on Facebook. We have temporarily removed the og:url property, but it would be nice to have this work again. 

**Expected behavior/output**

URL for the og:url property:
https://www.cadlab.tuc.gr/index.php?id=12493&tx_news_pi1%5Bnews%5D=19491&tx_news_pi1%5Bcontroller%5D=News&tx_news_pi1%5Baction%5D=detail&cHash=a61639f720d6b1bc0fb2e46d619276b5

**Environment**
- TYPO3 version(s): 9.5.19
- news version: 8.3.0
- Is your TYPO3 installation set up with Composer (Composer Mode): no
- OS: Ubuntu 18.04

**Possible Solution**

**Additional context**

Best regards
George Fiotakis

georgringer:
I switched to the API of the core which also just applies `htmlspecialchars` but I suggest having proper routing and it will be fine. If it feels still valid to you, please open an issue at https://forge.typo3.org/projects/typo3cms-core/issues

PR #1368, opened 2020-10-15
[BUGFIX] Fix meta og:url encoding (#1305)

seb37800:
Use third argument at false to avoid escape in the tag meta og:url


georgringer:
using the new metatag api now, therefore closing the issue

PR #1370, opened 2020-10-15
[TASK] Rename typoscript files with new .typoscript extension

simonschaufi:
Since TYPO3 8.7 the new file extension `.typoscript` is supported and should be used.

See https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.7.x/Feature-78161-IntroduceTypoScriptFileExtension.html


georgringer:
I know but this is a breaking change without much benefit besides having a different file ending

simonschaufi:
Can you explain why this is breaking? The core will just fetch the other files when you update the extension.

georgringer:
Because many integrators, including me, don't habe any sys_template record and include those files in their site extension manually

simonschaufi:
makes sense. Could you then please just fix the build by removing the trailing comma?

georgringer:
fixed with 6fcd2975ac29e0c7c7b47e4521df997abeb03b90

Issue #1376, opened 2020-10-29
Properties of type \DateTime have wrong annotations


christianbltr:
## Bug Report

Some properties in the models Link, News and TtContent have wrong annotations (type is declared as int instead of \DateTime) or they are missing.

I will open a pull request.

georgringer:
closing this as this would be quite breaking. If you need the datetime, you can easily transform it yourself. while I agree that datetime is nicer, there is no real benefit which would allow such breaking change

christianbltr:
The properties crdate, tstamp, starttime and endtime are already declared as type DateTime in the Models "Link", "News" and "TtContent" (example: https://github.com/georgringer/news/blob/master/Classes/Domain/Model/News.php#L19-L26).

The patch only fixes the annotations of the getters and setters to match to type declaration of the properties. So I would not consider this a breaking change.

I stumbled upon this because PhpStorm stated that I was passing variables with a wrong type when in fact I passed the correct one but only the annotation is wrong.

PR #1377, opened 2020-10-29
[TASK] Fix annotations

christianbltr:
fixes https://github.com/georgringer/news/issues/1376


georgringer:
closing this as this would be quite breaking. If you need the datetime, you can easily transform it yourself. while I agree that datetime is nicer, there is no real benefit which would allow such breaking change

georgringer (approved):

Issue #1425, opened 2021-01-18
MetaTag ViewHelper always uses attribute 'property'


cundd:
## Bug Report

**Current Behavior**
```
Input:
<n:metaTag property="something" content="{newsItem.title}" />
<n:metaTag name="something-else" content="{newsItem.title}" />
Output:
<meta property="something" content="TYPO3 is awesome" />
<meta property="something-else" content="TYPO3 is awesome" />
```

**Expected behavior/output**

I assume the intension of the different ViewHelper arguments `property` and `name` are that the meta-attribute can be changed. At the moment `property` is used in the call to `$pageRenderer->setMetaTag()` in both cases.

```
Input:
<n:metaTag property="something" content="{newsItem.title}" />
<n:metaTag name="something-else" content="{newsItem.title}" />
Output:
<meta property="something" content="TYPO3 is awesome" />
<meta name="something-else" content="TYPO3 is awesome" />
```

**Environment**
- TYPO3 version(s): 9.5.23
- news version: 8.5.2
- Is your TYPO3 installation set up with Composer (Composer Mode): no

georgringer:
resolved

PR #1426, opened 2021-01-18
[BUGFIX] Allow to use 'name' as attribute for meta-tag (#1425)

cundd:


christophlehmann:
Ah, made the same bugfix commit today :-) IMHO it's fine like this.

If one would improve it: It's also possible to use one meta tag with name and  property attribute, for example: `<meta name="title" property="og:title" content="xy"/>`.

georgringer:
imo fixed now already, closing this one

PR #1436, opened 2021-02-07
[BUGFIX] Fix SQL definition for related_link

### LINKED ISSUE #733: Avoid PHP warning with the "related_links"
## Bug Report

**Current Behavior**
When adding a news record in the backend and just save and close, in the DB the value stored for the field "related_links" is "NULL" (because it is a tinytext, I think for backwards compatibility)

After this, if I call the method "getRelatedLinks", I got a warning because the method return "NULL" instead of an instance of "LazyObjectStorage".

**Expected behavior/output**
No PHP warning...

**Environment**
- TYPO3 version(s): 8.7.19
- news version: 7.0.5
- OS: Ubuntu 16.04

**Possible Solution**
Just add a default value set to "0" in the TCA for the field "related_links".

Thanks

echavaillaz:
Resolves #733

### DIFF
--- a/ext_tables.sql
+++ b/ext_tables.sql
@@ -40,7 +40,7 @@ CREATE TABLE tx_news_domain_model_news (
 	related_from int(11) DEFAULT '0' NOT NULL,
 	related_files tinytext,
 	fal_related_files int(11) unsigned DEFAULT '0',
-	related_links tinytext,
+	related_links int(11) DEFAULT '0' NOT NULL,
 	type varchar(100) NOT NULL DEFAULT '0',
 	keywords text,
 	description text,


georgringer (approved):

mbrodala:
Due to this change I see the following when deploying the update to a system where the `tx_news_domain_model_news.related_links` field contains `NULL`:

```
Data truncated for column 'related_links' at row 5
```

Manual fix:

```sql
UPDATE tx_news_domain_model_news
SET related_links = 0
WHERE related_links IS NULL
```

georgringer:
@mbrodala should I change something?

mbrodala:
If the consuming code doesn't break you could allow `NULL`.

But my comment was mostly meant as notice with a workaround/solution if others stumble upon this.

benwalch:
@georgringer maybe mention the migration of @mbrodala (thanks btw) here:
https://docs.typo3.org/p/georgringer/news/9.1/en-us/Reference/Changelog/9-1-0.html

simplychrislike:
Had to write an upgrade wizard. Currently I have no time for a PR so I'll leave it here for now.

```php
<?php

declare(strict_types=1);

namespace Vendor\ExtName\Updates;

use TYPO3\CMS\Core\Database\Connection;
use TYPO3\CMS\Core\Database\ConnectionPool;
use TYPO3\CMS\Core\Database\Query\QueryBuilder;
use TYPO3\CMS\Core\Utility\GeneralUtility;
use TYPO3\CMS\Install\Updates\UpgradeWizardInterface;

final class SetNewsRelatedLinksIntegerDefaultValue implements UpgradeWizardInterface
{
    public const IDENTIFIER = 'extName_setNewsRelatedLinksIntegerDefaultValue';
    private const TABLE_NEWS = 'tx_news_domain_model_news';
    private const FIELD_UID = 'uid';
    private const FIELD_RELATED_LINKS = 'related_links';
    private const RELATED_LINKS_DEFAULT_VALUE = '0';

    /**
     * @inheritDoc
     */
    public function getIdentifier(): string
    {
        return self::IDENTIFIER;
    }

    /**
     * @inheritDoc
     */
    public function getTitle(): string
    {
        return sprintf(
            'Set integer default value where field `%s`.`%s` is NULL.',
            self::TABLE_NEWS,
            self::FIELD_RELATED_LINKS
        );
    }

    /**
     * @inheritDoc
     */
    public function getDescription(): string
    {
        return <<<'EOD'
            Set value to 0 (zero) where it's currently NULL to prevent errors
            during database schema update.
            See https://github.com/georgringer/news/pull/1436.
            EOD;
    }

    /**
     * @inheritDoc
     */
    public function executeUpdate(): bool
    {
        $query = self::getQueryBuilderForNews();
        $query
            ->update(self::TABLE_NEWS)
            ->set(
                self::FIELD_RELATED_LINKS,
                self::RELATED_LINKS_DEFAULT_VALUE,
                true,
                Connection::PARAM_INT
            )
            ->where($query->expr()->isNull(self::FIELD_RELATED_LINKS))
            ->execute();
        return true;
    }

    /**
     * @inheritDoc
     * @throws \Doctrine\DBAL\Driver\Exception
     */
    public function updateNecessary(): bool
    {
        $query = self::getQueryBuilderForNews();
        return (bool) $query
            ->count(self::FIELD_UID)
            ->from(self::TABLE_NEWS)
            ->where($query->expr()->isNull(self::FIELD_RELATED_LINKS))
            ->execute()
            ->fetchOne();
    }

    /**
     * @inheritDoc
     */
    public function getPrerequisites(): array
    {
        return [];
    }

    /**
     * @return QueryBuilder
     */
    private static function getQueryBuilderForNews(): QueryBuilder
    {
        $query = GeneralUtility::makeInstance(ConnectionPool::class)
            ->getQueryBuilderForTable(self::TABLE_NEWS);
        $query->getRestrictions()->removeAll();
        return $query;
    }
}

```

PR #1438, opened 2021-02-08
[BUGFIX] Fix page permissions SQL clause in Page-Utility

infabo:


bernhardberger (approved):
Looks fine to me. 

This is needed, so we don't stumble upon this kind of bug: https://forge.typo3.org/issues/75912

PR #1501, opened 2021-03-25
Symfony Dependency Injection (DI)

AKaravas:
**Classes**
Most classes that called other classes either with objectManager or GeneralUtility have been migrated to DI and can be found on the __construct() function. 

**Services.php**
In order for the DI to work the calsses have to be set to public. This can be achieved by adding the public attribute in the Services.yaml but it costs time since everytime the extension is called, these classes will be validated. Instead in the Services.php the classes will only be validated if they are called therefore saves time. 

**Domain Models**
I needed to test the imports and Georg pointed me to another extension which i used. The import was successfull but the results brought some problems. The import, imports only the delivered values, so if there is no teaser, then the entry in the Database is set to null. With the old News model, this brought an error (function getTeaser awaits for string, null given) and it could not show the entry in the FrontEnd. To avoid that, i assigned a default value to every protected variable that needs such value

### DIFF
--- a/Classes/Backend/FormDataProvider/NewsFlexFormManipulation.php
+++ b/Classes/Backend/FormDataProvider/NewsFlexFormManipulation.php
@@ -133,10 +133,13 @@ class NewsFlexFormManipulation implements FormDataProviderInterface
 
     /**
      * NewsFlexFormManipulation constructor.
+     * @param EmConfiguration $emConfiguration
      */
-    public function __construct()
+    public function __construct(
+        EmConfiguration $emConfiguration
+    )
     {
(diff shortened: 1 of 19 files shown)


AKaravas:
The failed checks have to do with the DI but TYPO3 sets the Controller and Domain Models by default on public. Therefore the NewsController should not have brought problems. Maybe integrate that on the Tests?

georgringer:
> The failed checks have to do with the DI but TYPO3 sets the Controller and Domain Models by default on public. Therefore the NewsController should not have brought problems. Maybe integrate that on the Tests?

sorry don't really know what you mean, can you take care of the failing tests?

AKaravas:
Yes, i will try to find where it fails

PR #1510, opened 2021-04-30
Add style to nonewsfound

ste101:


georgringer:
I won't merge that as this just adds another configuration and the template is being change anyways by the integrator

PR #1511, opened 2021-05-13
Make news/tags-relation bi-directional

kitzberger:
Setting up the bi-directional relation enables:
* BE users to see all related news when editing a tag record and alter that list
* integrators to render related news when rendering a tag on a news single view


georgringer (changes requested):
Hi,

Please change the tca: exclude should be true and remove the max items as 99 is for sure too less

PR #1554, opened 2021-08-12
[TASK] Switch from SignalSlot to Event in Importer classes

codegm:
Usage of SignalSlot is also deprecated since TYPO3v10


georgringer:
is there some docs to adopt as well? would you want to take care

PR #1555, opened 2021-08-13
[BUGFIX] Respect categories from demand

develth:
Fixes issue that categories set by overwriteDemand get lost

### DIFF
--- a/Classes/Controller/CategoryController.php
+++ b/Classes/Controller/CategoryController.php
@@ -30,7 +30,7 @@ class CategoryController extends NewsController
             $demand = $this->overwriteDemandObject($demand, $overwriteDemand);
         }
 
-        $idList = explode(',', $this->settings['categories']);
+        $idList = is_array($demand->getCategories())?$demand->getCategories():explode(',', $demand->getCategories());
 
         $startingPoint = null;
         if (!empty($this->settings['startingpoint'])) {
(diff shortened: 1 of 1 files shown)


Classes/Controller/CategoryController.php
```diff
@@ -30,7 +30,7 @@ public function listAction(array $overwriteDemand = null)
             $demand = $this->overwriteDemandObject($demand, $overwriteDemand);
         }
 
-        $idList = explode(',', $this->settings['categories']);
+        $idList = is_array($demand->getCategories())?$demand->getCategories():explode(',', $demand->getCategories());
```
DavidBruchmann, line 33:
```suggestion
        $idList = is_array($demand->getCategories()) ? $demand->getCategories() : explode(',', $demand->getCategories());
```

DavidBruchmann: I never checked functionality, it's just about the syntax.

georgringer (approved):

Megafry:
Hi
I have a list with several categories, ex:
_Categories 1, Categories 2, Categories 3_

With the old code I could set on as active, ex: 
_Categories 1, **Categories 2**, Categories 3_

Unfortunately this is no longer possible because I only get one category, ex:
**Categories 2**

Could this option be enabled / disabled via TypoScript ?

georgringer:
@Megafry done with 7ae3e1c0

develth:
Hi,
so currently there is no way to overwrite categories within filter in e.g. an ajax request?

PR #1560, opened 2021-08-31
Fix subclass registration

iresults-tma:
Without the array key, the configuration for news type 0 would be overwritten. News with type 0 are not displayed anymore.


echavaillaz:
I can confirm this behaviour, for me the code of news should also be updated like this in `Configuration/Extbase/Classes.php`:

```php
\GeorgRinger\News\Domain\Model\News::class => [
    'subclasses' => [
        0 => \GeorgRinger\News\Domain\Model\NewsDefault::class,
        1 => \GeorgRinger\News\Domain\Model\NewsInternal::class,
        2 => \GeorgRinger\News\Domain\Model\NewsExternal::class,
    ]
]
```
Adding the key is more explicit here.

georgringer (approved):

Issue #1525, opened 2021-06-14
Control (no)index and (no)follow of each news


t3easy:
## Feature Request

**Is your feature request related to a problem? Please describe.**
If i set the `index this site` switch of a news detail page to noindex, it gets removed from the core seo sitemap but all news also habe noindex in the head.
If i set it to index, the detail page is in the pages sitemap without any news to display and ext:news sends 404 header because no news is given.
Google search console is unhappy because I send pages that send 404.

**Describe the solution you'd like**
I want to remove the detail page without a displayed news from the sitemap and set it to noindex but let search engine index the news.
Therefore each news should have (no)index and (no)follow settings that overrules the page settings.

ste101:
You could leave no_index = 0 an add this to the sitemap config:
```
            pages {
                config {
                    additionalWhere := appendString( AND uid != 54)
                }
            }
```
But I would also like to see no_index in the seo tab of news because there are some news that are not worth to get indexed because they are only temporary like 'new year greatings'.
In this case the no_index meta would be useful.

christophlehmann:
I do it like this

* Set no_index on detail pages
* Use in pages sitemap `additionalWhere = AND no_index=0`

This way, detail pages are not in the sitemap. And then activate index,follow for the news detail view

```
[traverse(request.getQueryParams(), 'tx_news_pi1/news') > 0]
    page.meta.robots = TEXT
    page.meta.robots {
        value = index,follow
        replace = 1
    }
[global]
```

Would you like add it in the documentation if it fits your needs?

ste101:
> Would you like add it in the documentation if it fits your needs?

You can find this already here:
https://docs.typo3.org/p/georgringer/news/8.5/en-us/AdministratorManual/BestPractice/Seo/Index.html#robots-allow-indexing-of-news-records-but-set-noindex-to-the-detail-page

christophlehmann:
Great. We don't to add it ;)

georgringer:
@t3easy should we close the issue?

t3easy:
I would leave it open to implement a per news (no)index and (no)follow setting.

georgringer:
I am closing this issue as I don't plan to implement it.

if someone needs it, feel free to sponsor it and contact me in private

Issue #1160, opened 2020-02-17
Support for canonical URLs to link to original news article when syndicating


mabolek:
## Feature Request

**Is your feature request related to a problem? Please describe.**
I recently syndicated an article from richardhaeser.com on typo3.org. Without the possibility to manually add a canonical URL, Google will punish us.

**Describe the solution you'd like**
A "External Canonical URL" field when creating or editing a news article, where I can paste the link to the external article I am syndicating. The content of the field will be inserted into a `rel="canonical"` meta tag in the page header.

**Describe alternatives you've considered**
I have considered writing an extension extending News. However, I think this is a feature that could be useful to more people than myself. I'm therefore first trying to have it accepted as a new feature.

**Teachability, Documentation, Adoption, Migration Strategy**
Users will be able to use this new field with great ease. Educators (and the documentation) can instruct users to paste the URL to the original article in the "External Canonical URL" field. Users will feel happy and cared for. With the amount of unattributed second-hand content on the Internet, this will also serve as a reminder for editors to think before they paste.

georgringer:
this is something which should be done in a 3rd party extension - at least for the start. if you don't or can't create this extension, feel free to contact me

PR #1566, opened 2021-09-02
[FEATURE] Add robots (no_index, no_follow) and canonical

ste101:
This would solve #1525 and  #1160 .

Sitemap config in documentation should be ajusted with
`additionalWhere = no_index = 0 AND canonical_link = ''`
https://docs.typo3.org/p/georgringer/news/8.6/en-us/AdministratorManual/BestPractice/Seo/Index.html#basic-sitemap

What could be done better?


ayacoo:
Added a few comments for you from our branch.

Basically, Georg should decide whether we expand the EXT:news or whether it is then placed in the yoast_news.

georgringer:
this is something which should be done in a 3rd party extension

PR #1626, opened 2021-10-28
[FEATURE] permissions for Flexform settings

Gernott:
I added the exclude property, to make every Flexform value configurable in the BE-usergroup settings.


georgringer:
I am not a fan of polluting the permission module with tons of checkboxes. I never had the problem that only parts of the users have been allowed to a subset of fields.

because this is a breaking change I won't merge this and will close the PR

what about using tsconfig and hide fields for the user/group?

PR #1688, opened 2021-12-22
PHP Warning when show/hide news-plugin

TheNaderio:
Show/Hide a news-plugin in list-/page-module throws an Exception
```
PHP Warning: Trying to access array offset on value of type null in /var/www/html/public/typo3conf/ext/news/Classes/Hooks/BackendUtility.php line 138
```


linawolf (approved):
looks good to me

georgringer (approved):

PR #1735, opened 2022-03-04
[FEATURE] Filter for listing news recorders by be_user.

zikezhang:
show news recorder with limited by be_user.
So the current backend user can just see and edit his own records(and others if needed). Admins should see all records of course.

in theTSconfig of User Options tab:
options.delegatee = 3,4 // list id of be_user


zikezhang:
[I saw the request on stackoverflow](https://stackoverflow.com/questions/56931565/backend-access-rights-for-news-records/71212755#71212755)

georgringer:
Thanks for creating this issue. I won't merge that as the field cruser_id just holds the 1st editor id and is not able to be changed in backend

PR #1753, opened 2022-03-17
Update tests for category

froemken:


georgringer (changes requested):
The comments for getters and setters can be removed if only duplication of the method name. Same for parameters if the types are provided Already

georgringer (approved):

Issue #871, opened 2019-03-18
SearchForm Pagination


infabo:
## Bug Report

**Current Behavior**
Place Search Form Plugin. Place List Plugin below. Visit FE. Enter search word. See results. Need more than one result-pages. Navigate to page 2. Search-word lost - empty list.

**Expected behavior/output**
I want to see items of page 2.

**Environment**
- TYPO3 version(s): 8.7.24
- news version: 7.0.5
- Is your TYPO3 installation set up with Composer (Composer Mode): no
 OS: Debian

**Possible Solution**
https://github.com/georgringer/news/blob/master/Resources/Private/Templates/News/SearchForm.html defaults to a POST form.
The f:widget.link viewhelpers here https://github.com/georgringer/news/blob/master/Resources/Private/Templates/ViewHelpers/Widget/Paginate/Index.html have 'addQueryStringMethod="GET"'. It is clear how to fix, but this should not be broken out of the box.

**Additional context**
Add any other context about the problem here

jokumer:
I have written a solution to handle this feature in my own project/sitepackage extension.
If you are interested, i could share. But the best would be, to get it included in news itself.

p380:
hy,
the above code does not work on the latest version of the news module.

C0d3Br3aker:
hi,
is there now another solution for this problem?

jokumer:
https://github.com/georgringer/news/blob/master/Documentation/Misc/Changelog/9-0-0.rst#removed-viewhelpers

C0d3Br3aker:
@jokumer I need a solution for Version 8.6.0. I can't use 9.0.0 yet

PR #1805, opened 2022-06-13
[FEATURE] Transform search to GET

georgringer:
Skip not needed parameter and move the form to GET

Resolves: #871


xerc:
@ `typo3` 10.4.32 + `georgringer/news` 10.0.1

`Core: Exception handler (WEB): Uncaught TYPO3 Exception: Call to undefined method TYPO3\CMS\Fluid\Core\Rendering\RenderingContext::getUriBuilder() | Error thrown in file source/typo3conf/ext/news/Classes/ViewHelpers/SearchFormViewHelper.php in line 151`

PR #1812, opened 2022-06-14
Remove duplicate $configurationManager property

chesio:
Hi Georg,

I've noticed that the `$configurationManager` property is set in AdministrationController and NewsController, but since both of them inherit from abstract ActionController from Extbase, they inherit [this property](https://github.com/TYPO3/typo3/blob/main/typo3/sysext/extbase/Classes/Mvc/Controller/ActionController.php#L166) as well.

This PR removes the duplicate property declarations.


georgringer:
thanks for your PR! as the code changed slightly I did a new commit

Issue #1819, opened 2022-07-10
PHP8 V11: fetch() deprecated


aschmutt:
## Bug Report

**Current Behavior**
With a new installation of TYPO3 V11 with PHP8.1 and all Warnings enabled, I get these Exceptions, for example with News Administration Module in Backend and List view in Frontend. 

<img width="1221" alt="Bildschirmfoto 2022-07-10 um 17 25 58" src="https://user-images.githubusercontent.com/1215504/178151360-90539584-e085-4ea9-beab-59ed61535667.png">

PHP Runtime Deprecation Notice: PDOStatement::fetch(): Passing null to parameter #1 ($mode) of type int is deprecated in /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php line 138

<img width="1221" alt="Bildschirmfoto 2022-07-10 um 17 25 58" src="https://user-images.githubusercontent.com/1215504/178151379-a7c99474-6e3e-4518-8055-5f8546d5cd37.png">

**Expected behavior/output**
No Exception

**Environment**
- TYPO3 version(s): 11.5.12
- news version: 9.4.0
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: MacOS with ddev Setup

**Possible Solution**
replace all fetch() with fetchAssociative
This fixed my Exception:
`
class CategoryService, Line 103:

while (($row = $res->fetchAssociative())) {
`

**Additional context**
Searching in code showed more fetch() statements, so replacing all of them at once would be the best solution

aschmutt:
I will try to commit a patch

georgringer:
thanks for creating this issue. I will change the requirements a bit because such propose changed is not possible because of support of v10

aschmutt:
@georgringer I am not sure about moving to V12? 
I had a brand new V11 installation with more or less default settings in ddev, and even if it is just a deprecation warning, it blocks news completely in Backend and Frontend with this exception. 
Unfortunately, deprecations throw now an Exception in Dev Environment, until you deactivate this behaviour, so more people might get this: 
https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/Deprecation/Index.html

I used fetchAssociative already in V10 projects. I could not find out about fetchOne() and fetchAssociativeAll(), but if they are also available in V10, why not updating?

georgringer:
[...]

PR #1820, opened 2022-07-10
Bugfix/1819 Replace deprecated fetch functions

aschmutt:
[BUGFIX] replace deprecated fetch functions: 
- replace deprecated fetch() with fetchAssociative()
- replace deprecated fetchColumn() with fetchOne()
- replace deprecated fetchAll() with fetchAllAssociative()

Resolves: #1819


georgringer:
as v10 is still supported I need to change that a bit differently

PR #1824, opened 2022-07-20
[BUGFIX] extensionSummary Hook Call

lucmuller:
Whole params array should be sent to userfunction.


lucmuller:
Hello georg.

Thanks for merging the fix.
Would you mind releasing a news version of news extension as it is impacting one of our project using composer ?

georgringer:
sorry won't do that but just use composer patches to apply this oneliner

PR #1833, opened 2022-07-28
Propose simplified wording for timeRestriction

ursbraem:
I always stumble over these two fields in the plugin (I use the german translation). I usually just enter "today" and see what it does, then pick the desired one.  But it would be helpful also for editors to have a simpler wording.

If you agree on the simplified wording, I could also adapt it in german, but I haven't found ext_news on https://translation.typo3.org/de/ , where are the german translations (Oberes Zeiteinschränkung) stored?

### DIFF
--- a/Resources/Private/Language/locallang_be.xlf
+++ b/Resources/Private/Language/locallang_be.xlf
@@ -130,10 +130,10 @@
 				<source>Include subcategories</source>
 			</trans-unit>
 			<trans-unit id="flexforms_general.timeRestriction" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


georgringer:
the old translation server is not really relevant anymore since a long time, pls use https://crowdin.com/project/typo3-extension-news

georgringer:
what about using "News" or "Article" instead of "item"?

ursbraem:
Crowdin: nice! I've proposed translations for D / F / I / ES

PR #1834, opened 2022-07-31
[BUGFIX] PHP 8.1 warning in XML Sitemap DataProvider

peterkraume:
Undefined array key "useCategorySinglePid"


georgringer:
Please use?? False as otherwise also the value 0 would get true

PR #1836, opened 2022-08-01
[BUGFIX] Top panel buttons in BE module don't respect PageTS

kitzberger:
It's a fix for version `8.6.0` that I couldn't find a branch for ;-(


georgringer:
thanks. I am about to split the backend module up and change is integrated already

PR #1850, opened 2022-09-15
[FEATURE] Pre and post import eventDispatchers

### LINKED ISSUE #1849: Pre and post import eventDispatchers
How to keep news items in sync with an external source?

For example by deleting all previous records of a certain pid.

In the NewsImportService class there are 2 eventDispatchers which fire before processing a record and after. But there is no way to do custom stuf before or after the whole import process.

So, my question is can I propose code to add 2 additional eventDispatchers in the import function?

Zillion01:
Resolves #1849

### DIFF
--- a/Classes/Event/NewsPostImportEvent.php
+++ b/Classes/Event/NewsPostImportEvent.php
@@ -14,14 +14,8 @@ use GeorgRinger\News\Domain\Service\NewsImportService;
  */
 final class NewsPostImportEvent
 {
-    /**
-     * @var NewsImportService
-     */
     private $newsImportService;
 
-    /**
-     * @var array
-     */
     private $importData;
 
     public function __construct(NewsImportService $newsImportService, array $importData)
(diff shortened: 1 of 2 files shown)


Classes/Event/NewsPostImportEvent.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
```
georgringer, line 2: please add `declare(strict_types=1);`

Classes/Event/NewsPreImportEvent.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
```
georgringer, line 2: please add `declare(strict_types=1);`

Classes/Event/NewsPostImportEvent.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    private $importData;
+
+    public function __construct(NewsImportService $newsImportService, array $importData)
+    {
+        $this->newsImportService = $newsImportService;
+        $this->importData = $importData;
+    }
+
+    /**
```
georgringer, line 31: all those comments can be removed, also in other event

georgringer (changes requested):

Zillion01:
Thanks a lot

georgringer:
thanks to you ;)

Issue #1855, opened 2022-09-23
Adjustment of Pagination - after dispatching "NewsListActionEvent"


rasgor:
By using the NewsListActionEvent to post-process the newsRecords the pagination doesn't behave correctly.

Is it possible to move the pagination block from NewsController.php line 230 behind the event dispatching - NewsListActionEvent line 274?

georgringer:
thanks for creating the issue. wanna test https://github.com/georgringer/news/pull/1856?

rasgor:
Thank you for your fast response and adoption. After testing there is one last issue. The `$event->getAssignedValues()['news']` must be the param (basis) for the creation of the paginator ` $paginator = GeneralUtility::makeInstance(QueryResultPaginator::class, $event->getAssignedValues()['news'], $currentPage, $itemsPerPage, (int)($this->settings['limit'] ?? 0), (int)($this->settings['offset'] ?? 0));`

georgringer:
wanna do a PR?

rasgor:
Yes, later today.

georgringer:
I guess we can close it now

PR #1856, opened 2022-09-23
[BUGFIX] Move pagination after event #1855

georgringer:


rasgor (approved):
Looking good. Tested with my local dev env.

Issue #1868, opened 2022-10-07
access array offset in Workspaces


chriskrj:
## Bug Report

**Current Behavior**
When my Backend User is logged in and not in Live Workspace, Frontend breaks with this Error Message:

```
PHP Warning: Trying to access array offset on value of type null in /vendor/georgringer/news/Classes/ViewHelpers/LinkViewHelper.php line 257
```

**Expected behavior/output**
Show current Staging Workspace in Frontend

**Environment**
- TYPO3 version(s): [11.5.16]
- news version: [10.0.1]
- php: [8.1.1]
- Is your TYPO3 installation set up with Composer (Composer Mode): [yes]
- OS: [e.g. OSX 12.6]

**Possible Solution**
When i add isset in line 257 around $record[$uid] everything works fine again.
```php
if (isset($record['uid'])) {
  $uid = $record['uid'];
}
``` 

**Additional context**
Same behavior in news Version 9.4
I didn't check all View Helpers in TYPO3 11 / php 8.1 with Workspaces yet. But at the moment I do not have any other Errors with news and Workspaces.

PR #1872, opened 2022-10-11
[BUGFIX] - Update LinkViewHelper.php

chriskrj:
fixed array offset in php 8.1 and workspaces

https://github.com/georgringer/news/issues/1868

### DIFF
--- a/Classes/ViewHelpers/LinkViewHelper.php
+++ b/Classes/ViewHelpers/LinkViewHelper.php
@@ -254,7 +254,7 @@ class LinkViewHelper extends AbstractTagBasedViewHelper
                 'tx_news_domain_model_news',
                 $newsItem->getUid()
             );
-            if (isset($record['uid'] ?? false)) {
+            if (isset($record['uid'])) {
                 $uid = $record['uid'];
             }
         }


Classes/ViewHelpers/LinkViewHelper.php
```diff
@@ -254,7 +254,7 @@ protected function getNewsId(News $newsItem): int
                 'tx_news_domain_model_news',
                 $newsItem->getUid()
             );
-            if ($record['uid'] ?? false) {
+            if (isset($record['uid'] ?? false)) {
```
sbuerk, line 257: That does not make much sense. `??` is already the fallback if uid is not set in $record. Switching to isset() makes the fallback obsolute ... this does the same in two syntax variants.

sbuerk: It does not make sense to change it at all ... before and after doing the same check, just with different code variants.

Maybe the "issue" has existed before "?? false" fallback has been added ? Please retest it with the ?? fallback.

See following as demonstration why there is "no" need to change it here from ?? fallback to the isset variant:

https://3v4l.org/lu9ob

sbuerk: This change is superflous. It does chage the code but do nit fix anything or chage anything in behaviour. It can be closed.

sbuerk: or to make clear - $record['uid'] ?? false already covers the array key access / offset issue mentioned in the pull request message ...

See: https://3v4l.org/tU3NG

sbuerk: look - you should updatethe version. You get the error because you do not have the "??" fallback. 

The issue is already fixed in the code.

You should just update to get it. Your PR came to late. That is what I mean - you are change an already existing fix for your issue to another variant, thus making it superflous.

sbuerk:
This PR can be cloes. The "issue" which this PR addresses is already fixed with

https://github.com/georgringer/news/commit/307e1c00390000c13734d08367206745108a38cc

and released as 10.0.2

PR #1893, opened 2022-10-25
[TASK] Add invidiual icons for content elements

malteriechmann:
@georgringer asked for some help [on Twitter](https://twitter.com/georg_ringer/status/1583420546318217216).

After a quick chat, I asked one of our designers to create some icons and here we are. Will those fit your needs?

**Important:** I just changed those icons directly in source code and did not try to run this extension with these changes. Should work properly though. If this was a stable branch, I would make manual tests for sure.


georgringer:
Thanks a lot!
I think the searchForm/-Result need some more love.
- the magnifier belongs IMO more to the searchform
- the search result needs maybe magnifier + parts of the list?
what do you think?

malteriechmann:
Sounds good and I agree. I will come back to you.

What do you think about the list icon (with and without fallback)? Does it makes sense for you?

georgringer:
thanks! for people with color blindness the searchForm and searchresult icons can't be distinguished. I would suggest to remove the lines in the searchform as this action does only render the search form and no result. then it would be perfectly fine!

malteriechmann:
We updated the search form icon. What do you think?

georgringer:
i like it!

PR #1982, opened 2023-01-08
DateMenu.html throws exception

codegm:
(1/1) #1280825466 TYPO3\CMS\Extbase\Exception
There is more than one plugin that can handle this request (Extension: "News", Controller: "News", action: "list"). Please specify "pluginName" argument

Branch: 11-12

### DIFF
--- a/Resources/Private/Templates/News/DateMenu.html
+++ b/Resources/Private/Templates/News/DateMenu.html
@@ -22,7 +22,7 @@
 									<li class="item">
 								</f:else>
 							</f:if>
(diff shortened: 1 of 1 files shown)


georgringer:
thanks for the issue and testing the branch! IMO this should be fixed in the core and not in news itself. I created the issue https://forge.typo3.org/issues/99501 and the affected change https://review.typo3.org/c/Packages/TYPO3.CMS/+/77294

let's see if this will be merged

georgringer:
FYI I xlcassed it now in ext:news as it is still not merged in core :/

Issue #1984, opened 2023-01-09
Loading Issue with large category tree


abohntek:
## Bug Report

**Current Behavior**
for dozens of categories the tree takes up to 15  seconds to be loaded. If the news is saved before the "Navigation loading error." error appears. For Typo3 core (pages, tt_content) the error is fixed by fasting up the form engine: https://forge.typo3.org/issues/97003 for news this behaviour still happens.

**Expected behavior/output**
Category tree is as fast loaded as for pages and tt_content elements. 

**Environment**
- TYPO3 version(s): 11.5.21
- news version: 10.0.3
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: Ubuntu 24

georgringer:
thanks for the issue. the used code for the tree should be in theory the same?! wanna create a PR or sponsor the fix of the issue?

georgringer:
solved, closing

PR #1985, opened 2023-01-09
Loading Issue with large category tree

abohntek:
#1984


georgringer:
you need to add the change here: https://github.com/georgringer/news/blob/main/Configuration/TCA/Overrides/z_misc_11_adoptions.php as the feature is only available since 11 but not in 10!

georgringer (approved):

PR #1990, opened 2023-01-20
Slash missing

ste101:


georgringer:
which issue does occur when the slash is missing?

simonschaufi:
I think it's more about consistency than an actual error

PR #2005, opened 2023-02-16
typo3/cms-composer-installers 3.1.3 for PHP 8.2 compatibility

liayn:


sbuerk:
This change dissallowed needed ^5 composer installers needed for TYPO3 v12. That has been undetected, that the tests for v12 has been installed v11 instead (due root composer.json conflict).

Issue #2009, opened 2023-02-17
Error when running cleanup:deletedrecords / category label hook called without db row


akiessling:
I get an error during when i run the cleanup command. Somehow the hook is called without a row and then fails here. I added a var_dump right before and $row is just null

`vendor/bin/typo3cms cleanup:deletedrecords`

https://github.com/georgringer/news/blob/f4e0c3f68b41eace1f7cc91389475942f9cf5702/Classes/Hooks/Labels.php#L42

```
Deletion process starting now.
------------------------------

Flushing 53 deleted records from table "sys_category"
array(4) {
  ["table"]=>
  string(12) "sys_category"
  ["row"]=>
  NULL
  ["title"]=>
  string(0) ""
  ["options"]=>
  array(0) {
  }
}

                                                                
  [ ErrorReporting\Warning ]                                    
  Warning: Trying to access array offset on value of type null  
```

```
Exception trace:
#0 ()
   typo3conf/ext/news/Classes/Hooks/Labels.php:42
#1 GeorgRinger\News\Hooks\Labels->getUserLabelCategory()
#2 call_user_func_array()
   typo3/sysext/core/Classes/Utility/GeneralUtility.php:3126
#3 TYPO3\CMS\Core\Utility\GeneralUtility::callUserFunction()
   typo3/sysext/backend/Classes/Utility/BackendUtility.php:1431
#4 TYPO3\CMS\Backend\Utility\BackendUtility::getRecordTitle()
   typo3/sysext/core/Classes/DataHandling/DataHandler.php:7280
#5 TYPO3\CMS\Core\DataHandling\DataHandler->getRecordPropertiesFromRow()
   typo3/sysext/core/Classes/DataHandling/DataHandler.php:7264
#6 TYPO3\CMS\Core\DataHandling\DataHandler->getRecordProperties()
   typo3/sysext/core/Classes/DataHandling/DataHandler.php:5053
#7 TYPO3\CMS\Core\DataHandling\DataHandler->deleteRecord()
   typo3/sysext/lowlevel/Classes/Command/DeletedRecordsCommand.php:315
```
                                                                
TYPO3 11.5
PHP 8.1
News 10.0.3

georgringer:
wanna do a PR and just continue/early return if not there?

akiessling:
Done. I wrapped it all in an isset since the method has no return value

PR #2012, opened 2023-02-19
[BUGFIX] prevent error in cleanup:deletedrecords command

akiessling:
The category label hook might get called without a record during the cleanup and stops the execution of the cleanup command.

Fixes: #2009


georgringer (changes requested):
I suggest an !isset and an early return;

akiessling:
> I suggest an !isset and an early return;

done

akiessling:
Thanks!

PR #2066, opened 2023-04-06
Use `if` with DisableLanguageMenuProcessor

julianhofmann:
DisableLanguageMenuProcessor should only be processed if a news-record is given, else it is struggling with a wrong object for the route and end up in `Call to undefined method TYPO3\CMS\Core\Routing\SiteRouteResult::getRouteArguments()`. Another option would be inserting the snippet *only* on the detail page.


julianhofmann:
The addition of the documentation is appropriate in terms of content, but at the moment the `if` is unfortunately ignored. While many DataProcessors support the `if` property, the DisableLanguageMenuProcessor still lacks this.

PR is following soon.

PR #2096, opened 2023-04-27
[Task] Improved ObjectStorage return type with generic type parameter

KrohnMi:


Classes/Domain/Model/News.php
```diff
@@ -543,7 +543,7 @@ public function getRelated()
     /**
      * Set related from
      *
-     * @param ObjectStorage $relatedFrom
+     * @param ObjectStorage<\GeorgRinger\News\Domain\Model\Category> $relatedFrom
```
georgringer, line 546: shouldn't that be news model?

georgringer (changes requested):

KrohnMi:
@georgringer Can you please check again.

PR #2097, opened 2023-04-28
[Bugfix] FlexForm setting ItemsPerPage in SearchResult works now 

markuskappe:


georgringer:
fixed a different way with b7bb5547b44eacc1e280ffe752d3dbbcd649bd34

PR #2116, opened 2023-05-22
[BUGFIX] Uncaught TypeError: trim() expects parameter 1 to be string, null given

hmccloy:
Thrown in typo3conf/ext/news/Classes/Updates/PluginUpdater.php on line 181


georgringer:
IMO this is fixed, closing for now

PR #2128, opened 2023-06-19
[BUGFIX] Create proper canonical URLs for the single view

oliverklee:
URLs for the single view usually have `action`, `controller` and `news` parameters. So we will need to include all of those into the list of canonicalized URL parameters in order to have proper canonical URLs even if the original URLs does not use the slug yet.


oliverklee:
The CGL failures do not seem related to my changes.

oliverklee:
I have created #2129 for the CS fixes.

georgringer (approved):

Issue #2172, opened 2023-08-17
Plugin updater removes additional flexform fields added via hooks


eliashaeussler:
## Bug Report

**Current Behavior**
We've extended the news flexform using a dedicated hook for `FlexFormTools`, registered within `$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS'][\TYPO3\CMS\Core\Configuration\FlexForm\FlexFormTools::class]['flexParsing']`. This works fine and the additional fields are shown in FormEngine.

However, when running the `PluginUpdater` upgrade wizard, those fields get removed from the `pi_flexform` values. This is because of the following method: https://github.com/georgringer/news/blob/a81398060ff206bc7c7c1aff4b148e2ceadc7499/Classes/Updates/PluginUpdater.php#L185

It obviously does not include additional fields added via hooks and therefore removes all of them.

**Expected behavior/output**
It would be nice if additional fields registered via flexform hooks can be taken into account when cleaning up flexform XMLs. This would avoid additional fields from getting removed, leading to a potential change in behavior (depending on what additional fields are configured and what impact they have).

**Environment**
- TYPO3 version(s): 11.5.30
- news version: 11.1.2
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: macOS 13.4.1

**Possible Solution**
I have a working solution in place which uses `TYPO3\CMS\Core\Configuration\FlexForm\FlexFormTools::cleanFlexFormXML()` to clean up the flexform XML. AFAICS, this method does a pretty similar job like `GeorgRinger\News\Updates\PluginUpdater:: getAllowedSettingsFromFlexForm()`, but additionally calls all registered hooks. I'm not sure if this solution could be used as drop-in replacement for the current implementation, therefore I'm posting it here before submitting a PR. Please let me know if this is something you'd consider useful and I'd be happy to submit a PR.

georgringer:
thanks for the issue. yes please add the PR

eliashaeussler:
Alright, here we go: #2173

liayn:
Can confirm with eventnews

PR #2173, opened 2023-08-18
[BUGFIX] Use FlexFormTools to clean up flexforms in plugin updater

eliashaeussler:
By using the `FlexFormTools->cleanFlexFormXML()` method, registered hooks for flexform modifications are also taken into account. Without its usage, those additional fields would get removed during an upgrade.

Resolves: #2172


liayn:
I justed tested dev-main again, but unfortunately this now leads to:
`PHP Warning: Undefined global variable $LANG in /var/www/html/vendor/typo3/cms-form/Classes/Hooks/DataStructureIdentifierHook.php line 403`

So we have a hook in Core, that assumes `$GLOBALS['LANG']` is present.
What shall we do? Init $LANG? Fix Core?

DavidBruchmann:
It's very unlikely that `$GLOBALS['LANG']` will be re-introduced in the core.

liayn:
> It's very unlikely that `$GLOBALS['LANG']` will be re-introduced in the core.

Not sure what you mean? I'm testing on v11 here, so Core is **using** it by itself.
The fix in v11 Core would be to not rely on the global an fallback to creating a dedicated instance of the `LanguageService`.

Besides that: The fix does not work with eventnews v6 currently. I'm still investigating this.

liayn:
Note: 

> What shall we do? Init $LANG? Fix Core?

Initializing `$GLOBALS['LANG']` in the PluginUpdater helps to mitigate the problem in v11 at least.

> Besides that: The fix does not work with eventnews v6 currently. I'm still investigating this.

Something is wrong with the generated DS somehow.

DavidBruchmann:
anyway, it's then an issue for the form extension and not directly related to news.

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.0/Deprecation-88567-GLOBALS_LOCAL_LANG.html?highlight=globals%20lang

liayn:
> anyway, it's then an issue for the form extension and not directly related to news.
> 
> https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.0/Deprecation-88567-GLOBALS_LOCAL_LANG.html?highlight=globals%20lang

` $GLOBALS['LOCAL_LANG']`  vs. ` $GLOBALS['LANG']` :-)

It is an issue of news here too, because it is using the Flexform stuff outside the expected context. The ext-form hook never expected to be called in an install-tool-context.
So, I'd say we have to cope with this in news.

DavidBruchmann:
concerning the DS Elias provided just a PR which is already merged I think, not sure if this could be related

DavidBruchmann:
ahh this change :/

liayn:
The ext:form issue is solved with: https://github.com/georgringer/news/pull/2197

IchHabRecht:
For me the creation of LanguageService failed in a current TYPO3 12.4 installation. I added another pull request: https://github.com/georgringer/news/pull/2199

PR #2189, opened 2023-09-06
[TASK] Remove redundant ORDER BY clause in default_sortby

nhovratov:
The ORDER BY clause is redundant and is removed by TYPO3 anyway.

The link type defined both, sortby and default_sortby. This is invalid. In this case sortby overrules default_sortby and hence the latter can be removed.


nhovratov:
For reference: https://docs.typo3.org/m/typo3/reference-tca/main/en-us/Ctrl/Properties/DefaultSortby.html

PR #2191, opened 2023-09-13
[TASK] Typo in RSS Feed TypoScript Example

SvenJuergens:
should be tt_content.list.20.news_pi1 instead of tt_content.20.news_pi1


georgringer:
IMO it should be `tt_content.news_pi1.20`  check the object browser, shouldn't it?

SvenJuergens:
hi @georgringer 
you are right, sorry in News 11 it is this path, I still had an older version installed, there it is still the other path and I have not set in the documentation in the TER to the correct version. Sorry, I'll close the PR here again :)

PR #2199, opened 2023-09-18
[BUGFIX] Correctly create LanguageService in UpgradeWizard

IchHabRecht:
As the LanguageService has own constructor arguments, the Factory needs to be used to create a valid LanguageService object.


georgringer:
is this also compatible to v11?

PR #2209, opened 2023-10-09
[CLEANUP] Autoformat the code with the latest PHP-CS-Fixer version

oliverklee:


oliverklee:
It looks like StyleCI and PHP-CS-Fixer do not agree with each other. I propose ditching StyleCI and only use the Core PHP-CS-Fixer ruleset.

oliverklee:
This PR is blocked by #2210.

Issue #2214, opened 2023-10-09
Functional test failures with 12LTS with highest dependencies


oliverklee:
The 11LTS tests are green both with the highest and the lowest dependencies:
https://github.com/georgringer/news/actions/runs/6461252353

The 12LTS tests are green with the lowest dependencies, but are failing with the highest dependencies:
https://github.com/georgringer/news/actions/runs/6461252351

An example:
``` 
3) GeorgRinger\News\Tests\Functional\Repository\NewsRepositoryTest::findTopNewsRecords
Doctrine\DBAL\Exception\UniqueConstraintViolationException: An exception occurred while executing a query: Duplicate entry '134-5' for key 'PRIMARY'

/home/runner/work/news/news/.Build/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php:62
/home/runner/work/news/news/.Build/vendor/doctrine/dbal/src/Connection.php:1938
/home/runner/work/news/news/.Build/vendor/doctrine/dbal/src/Connection.php:1880
/home/runner/work/news/news/.Build/vendor/doctrine/dbal/src/Connection.php:1208
/home/runner/work/news/news/.Build/vendor/doctrine/dbal/src/Connection.php:808
/home/runner/work/news/news/.Build/vendor/typo3/cms-core/Classes/Database/Connection.php:202
/home/runner/work/news/news/.Build/vendor/typo3/testing-framework/Classes/Core/DatabaseConnectionWrapper.php:59
/home/runner/work/news/news/.Build/vendor/typo3/testing-framework/Classes/Core/Functional/FunctionalTestCase.php:717
/home/runner/work/news/news/Tests/Functional/Repository/NewsRepositoryTest.php:42
```

georgringer:
IMO fixed

PR #2215, opened 2023-10-09
[TASK] Only allow LTS core versions

oliverklee:
For TYPO3 12, the minimum version should be the LTS release 12.4.0, not any lower version like 12.2 or 12.3.


oliverklee:
The failures on 12LTS with the highest dependencies most probably are not related to this change. I've created a ticket for that problem: https://github.com/georgringer/news/issues/2214

PR #2271, opened 2023-11-22
[FEATURE] Allow to set a ttl for the rss feed

KrohnMi:


georgringer:
isnt there a change missing in the setup.typoscript?

KrohnMi:
You're right, I've added.

Issue #2272, opened 2023-11-23
Proxy Class Generation: extending contructor not working for tags


kitzberger:
## Bug Report

**Current Behavior**
When extending a model class and using a `__construct()` method in the own class is only working when that method exists in original class, e.g. `Tag`.

There's no real exception thrown though.

**Expected behavior/output**
Constructor can be set in own model class.

**Environment**
- TYPO3 version(s): 11
- news version: 11
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: linux

**Possible Solution**
Maybe add an empty constructor to all model classes?

georgringer:
wanna add a PR with empty constructor + a comment that this is intentional?

PR #2274, opened 2023-11-24
[BUGFIX] Make sure all model classes are extensible (#2272)

kitzberger:


kitzberger:
@georgringer, the one remaining failed test suggests to use the code style the other tests didn't like :joy:

PR #2290, opened 2023-12-05
[TASK] Change method visibility

KrohnMi:
I'm having trouble understanding why the methods were set as protected. The class is declared as final, which means it cannot be extended. Additionally, I don't see any calls to these methods within the class itself.


georgringer:
derived from `abstract class AbstractPaginator` currently I don't see a reason why changing it

PR #2301, opened 2023-12-30
[DOCS] Add start anchor to the index page

linawolf:
This is a defined standard in the Documentations Team Recommendations


linawolf:
@georgringer I don't think anything in this PR could have caused the test failures, there must be some other issue

PR #2315, opened 2024-01-16
[BUGFIX] Make `symfony/console` an explicit dependency

oliverklee:
Relying on transitional dependencies (and their version constraints) is considered bad practice.

Related to #2312

PR #2328, opened 2024-01-17
[TASK] Properly define dependencies on system extensions

oliverklee:


georgringer:
Looks good, what about adding scheduler to suggest as well?

oliverklee:
Marking as draft. I'd like #2315 and #2329 to get merged first.

oliverklee:
> what about adding scheduler to suggest as well?

Done.

PR #2342, opened 2024-01-28
Improve documentation

kanow:
Add RenderMediaViewHelper reference to the docs and fix sphinx warning with wrong url of t3extbasebook.


kanow:
Something seems wrong with older commits.

PR #2377, opened 2024-04-10
[BUGFIX] Undefined array key "categories" in CategoryController

80Quattro:


georgringer:
sorry but the array key is already checked using the ?? and I don't want the additional fallback to the demand!

PR #2458, opened 2024-04-22
[DOCS] Modernize outdated conditions still using globalVar

ErHaWeb:
With the introduction of the symfony expression language since TYPO3 9.4, the old syntax (such as the use of `globalVar`) is considered as deprecated. This patch converts all conditions to the new syntax.

A [null-safe operator in the TypoScript conditions](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.1/Feature-100586-Null-safeOperatorInTypoScriptConditions.html#feature-100586-null-safe-operator-in-typoscript-conditions) was deliberately not used in this patch, as this was only introduced from TYPO3 v12.1 onwards. As soon as support for v11 is ended, this more elegant abbreviation `[getTSFE()?.id == 9819]` can be used accordingly.

In addition, `[global]` is replaced by the more suitable variant `[END]`, which only terminates the previous condition.

### DIFF
--- a/Documentation/Tutorials/BestPractice/ICalendar/Index.rst
+++ b/Documentation/Tutorials/BestPractice/ICalendar/Index.rst
@@ -29,7 +29,7 @@ A very simple way to generate the iCalendar feed is using plain TypoScript. All 
 
 .. code-block:: typoscript
 
-    [globalVar = TSFE:type = 9819]
+    [getTSFE() && getTSFE().id == 9819]
     config {
        disableAllHeaderCode = 1
        xhtml_cleaning = none
@@ -60,7 +60,7 @@ A very simple way to generate the iCalendar feed is using plain TypoScript. All 
           }
        }
     }
-    [global]
+    [END]
 
 This example will show all news records which don't have the category with the uid 9 assigned and are saved on the page with uid 24.
 
@@ -121,7 +121,7 @@ The TypoScript code looks like this.
 
 .. code-block:: typoscript
 
-    [globalVar = TSFE:type = 9819]
+    [getTSFE() && getTSFE().id == 9819]
        lib.stdheader >
        tt_content.stdWrap.innerWrap >
        tt_content.stdWrap.wrap >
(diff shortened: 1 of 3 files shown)


Documentation/Tutorials/BestPractice/ICalendar/Index.rst
```diff
@@ -29,7 +29,7 @@ A very simple way to generate the iCalendar feed is using plain TypoScript. All
 
 .. code-block:: typoscript
 
-    [globalVar = TSFE:type = 9819]
+    [getTSFE() && getTSFE().id == 9819]
```
linawolf, line 32: I would suggest to use traverse() also id and type is not the same thing...

georgringer: @linawolf wanna do a PR for that?

ErHaWeb: @linawolf Thank you for the comment, of course you're right. I had taken this from the documentation and overlooked the fact that id is used there instead of type. I will create a new PR in which traverse and type is used.

ErHaWeb: @linawolf Which spelling of `traverse` do you mean exactly?

The notation `[traverse(page, 'type') == 9819]` does not work, because `type` is not a field of the `pages` table and therefore not in the current `page` array, which is used here and contains all record data of the current page.

I see no other way to retrieve the TSFE data or at least the current `type` of the called page than via `[getTSFE() && getTSFE().type == 9819]` (or from version 12 and higher then with the null-safe operator `[getTSFE()?.type == 9819]`).

The same problem has already been discussed here on Forge in [issue #100563](https://forge.typo3.org/issues/100563) with the result of using `getTSFE()` as I originally intended.

In my opinion, this is also the easiest to understand (at least for beginners). In the case of `traverse`, if it comes into question for this use case, you must first understand what this function does.

For these reasons, I will create my new PR with `[getTSFE() && getTSFE().type == 9819]`. Feel free to change this to a `traverse` notation if you know of one and feel this is more appropriate here.

ErHaWeb:
This PR was closed in favor of the newer PR #2469.

PR #2474, opened 2024-05-03
[BUGFIX] Fix record overlay for 'List of selected items' (2369)

### LINKED ISSUE #2369: The "List of selected items" view with "sort by" setting does not respect language overlay
## Bug Report

**Current Behavior**
Place a news plugin with view type "List of selected items" on a localized page.
Select default language of one or more single news records.
Select any "Sort by" option e.g. "Given date/time".

**Expected behavior/output**
Localized news should appear on frontend page but result is an empty page.

**Environment**
- TYPO3 version(s): 10.4
- news version: 9.4.0
- OS: Windows 11

**Possible Solution**
If "sort by" option is NOT set, uids of selected news records are passed to findByIdentifier() method of repository which seems to look in uid AND l10n_parent.
If "sort by" option is set, $query->in('uid', X) is used and localizations are missing.

Idleworks:
Fixes #2369

### DIFF
--- a/Classes/Domain/Repository/NewsRepository.php
+++ b/Classes/Domain/Repository/NewsRepository.php
@@ -251,6 +251,9 @@ class NewsRepository extends AbstractDemandedRepository
         // Id list
         $idList = $demand->getIdList();
         if ($idList) {
+            $query->getQuerySettings()->setRespectStoragePage(false);
+            $query->getQuerySettings()->setRespectSysLanguage(false);
+            $query->getQuerySettings()->setLanguageOverlayMode(true);
             $constraints['idList'] = $query->in('uid', GeneralUtility::intExplode(',', $idList, true));
         }
 


Classes/Domain/Repository/NewsRepository.php
```diff
@@ -251,6 +251,9 @@ protected function createConstraintsFromDemand(QueryInterface $query, DemandInte
         // Id list
         $idList = $demand->getIdList();
         if ($idList) {
+            $query->getQuerySettings()->setRespectStoragePage(false);
```
georgringer, line 254: this is not needed as already set in `generateQuery()`

Classes/Domain/Repository/NewsRepository.php
```diff
@@ -251,6 +251,9 @@ protected function createConstraintsFromDemand(QueryInterface $query, DemandInte
...
+            $query->getQuerySettings()->setRespectSysLanguage(false);
```
georgringer, line 255: why not set $disableLanguageOverlayMode to true in the selectedListAction call?

Classes/Domain/Repository/NewsRepository.php
```diff
@@ -251,6 +251,9 @@ protected function createConstraintsFromDemand(QueryInterface $query, DemandInte
...
+            $query->getQuerySettings()->setLanguageOverlayMode(true);
```
georgringer, line 256: I would also move that out of this important method which is used by **all** actions and make if configurable from the action. what do you think?

georgringer (changes requested):

PR #2478, opened 2024-05-14
[BUGFIX] Fix passed argument is array and not string

mai-space:


georgringer:
fixed already with another PR but thanks a lot!

PR #2480, opened 2024-05-17
[TASK] Streamline Build/Scripts/runTests.sh

sbuerk:
This change modifies the `Build/Scripts/runTests.sh`
execution wrapper to be in line with current TYPO3
core implementation.

This contains following tasks:

* Modify `Build/Scripts/runTests.sh`:
    * `-b <container-binary>` switch to determine which
      container binary should be used.
    * Remove the usage of `docker-compose.yml` file and
      usage by replacing container execution call with
      docker or podman binary calls.
    * Include `bash trap` logic for container/network
      cleanup on abort signals.
    * deprecated the `-e` option in favour of direct
      command expansion using `--`.
* Refactor suit command execution without docker-compose.
* Remove `Build/testing-docker/docker-compose.yml` due
  to being obsolete now.
* Add composer related temporary files to `.gitignore`
  and `.gitattributes`.
* Add documentation rendering and test rendering to the
  `runTests.sh`.
* Switched composer validate to remove deprecated `-e`
  option usage in GitHub action workflow and added the
  `--no-check-lock` option to mitigate lock file errors
  due to core switches.
* Add `-s docsGenerate` as suite to `Build/Scripts/runTests.sh`
  to render the documentation.


Build/Scripts/runTests.sh
```diff
@@ -164,149 +262,141 @@ Options:
...
+        New images are published once in a while and only the latest ones are supported by core testing.
+        Use this if weird test errors occur. Also removes obsolete image versions of typo3/core-testing-*.
 
     -h
         Show this help.
 
 Examples:
-    # Run unit tests using PHP 7.4
+    # Run all core unit tests using PHP 8.3
+    ./Build/Scripts/runTests.sh
```
CybotTM, line 279: wouldn't this use PHP 7.4 and not 8.3? as 7.4 being default.

sbuerk: correct. fixed.

linawolf (approved):
Thanks, I like this

Issue #2495, opened 2024-06-18
Filter news by categories with multiple select fields


kanow:
## Bug Report

**Current Behavior**
I'm adding a news plugin with no category restriction, to show all news from a sysfolder.
In fluid template I built two select fields for categories.
If I select only one category in the first select field, te result is empty. If I select categories in both of the select fields, then the result is not empty. It shows news data with selected categpories due to the category conjunction "and". The reason for that is an empty second category in demand object

```
categories => protected array(2 items)
         0 => '39' (2 chars)
         1 => '' (0 chars)
```

If I change categoryConjunction to "or" it works. But thats not what I want. I need categoryConjunction to "and" for the two categories to get the expected result.

**Expected behavior/output**
Get the news with the selected category in the search result also if only one select field was used to select a category.

**Environment**
- TYPO3 version(s): 12.4.14
- news version: 11.4.2
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: Local docker environment with PHP 8.2

**Possible Solution**
Use `array_filter` to remove empty values
Add this line after line 205 in NewsController in the `overwriteDemandObject` method.

```
$propertyValue = is_array($propertyValue) ? array_filter($propertyValue) : $propertyValue;
```

Maybe `array_filter` is not the best solution or we need.

georgringer:
maybe use explode + GeneralUtility::intExplode(',', $array, true)

PR #2496, opened 2024-06-18
[BUGFIX] Fix issue with empty category in overwriteDemand

kanow:
Related: #2495

If more than one category is used for a filter form, then empty categories (not selected in from) are still added to the category constraint. This PR will remove empty categories in overwriteDemand object.


Classes/Controller/NewsController.php
```diff
@@ -203,6 +203,7 @@ protected function overwriteDemandObject(NewsDemand $demand, array $overwriteDem
                     if (!is_array($propertyValue)) {
                         $propertyValue = GeneralUtility::trimExplode(',', $propertyValue, true);
                     }
+                    $propertyValue = is_array($propertyValue) ? GeneralUtility::intExplode(',', implode(',',$propertyValue), true) : $propertyValue;
```
georgringer, line 206: should be
```
$propertyValue = is_array($propertyValue) ? GeneralUtility::intExplode(',', implode(',', $propertyValue), true) : $propertyValue;
```

georgringer (changes requested):

Classes/Controller/NewsController.php
kanow: Yes, of course :grin:

Issue #2505, opened 2024-07-22
clearCachePostProc hook repeats calls


heifiDR:
## Bug Report

**Current Behavior**
When the clearCachePostProc() in the DataHandlerHook is called, it always writes 2 entries to the $cacheTagsToFlush array. It then loops through the array and executes $cacheManager->flushCachesInGroupByTag(). Because the class is a singleton, the $cacheTagsToFlush array is never deleted. This means that when another news item is deleted, the cache for all previous items is emptied again. 

With 115 deleted news elements, that are on the same page, the $cacheManager->flushCachesInGroupByTag() call occurs over 6600 times.

**Expected behavior/output**
The $cacheManager->flushCachesInGroupByTag() should just be called once per item and page.

**Environment**
- TYPO3 version(s): [11, 12]
- news version: [10.0.3, 11.4.2]
- Is your TYPO3 installation set up with Composer (Composer Mode): [yes]

**Possible Solution**
There are different ways to solve this. The fastest solution could be to reset the $cacheTagsToFlush array on each hook call, but this would still trigger multiple calls for the same page. The cleanest approach would probably be to collect alle $cacheTagsToFlush in an earlier hook and to clear it in the clearCachePostProc() Hook.

georgringer:
wanna provide a PR to improve the situation? last suggestion looks best

PR #2508, opened 2024-07-26
[BUGFIX] clear the flushed cache tags (#2505)

heifiDR:
After investigating how a solution could look like, I am back to the simple way. Unfortunately I found no hook which can be triggerd after foreaching through each element with prepareCacheFlush(). There is no clean way to get another usefull clear_cacheCmd out of the prepareCacheFlush() method. And there is no way to manipulate the $tagsToClear per hook, to let the core do the cache flushing with the custom tags.

With my change, already flushed cache tags get removed from the array. It's not perfect, because a page with multiple items on it is still flushed multiple times. But the solution avoids the linear increasing workload.


georgringer:
fixed with f0e94c88da5b07bd5a2cc566ca322484fd08617b, closing

Issue #2475, opened 2024-05-06
Introduce Site sets for TYPO3 13


linawolf:
## Feature Request
Introduce site sets, https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.1/Feature-103437-IntroduceSiteSets.html for TYPO3 13

PR #2519, opened 2024-08-15
[FEATURE] Introduce site sets for v13

linawolf:
See also https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/SiteHandling/SiteSets.html

SiteSets

Resolves https://github.com/georgringer/news/issues/2475


linawolf:
Depends on #2520

Configuration/TypoScript/SeoSitemap/setup.typoscript
```diff
@@ -17,10 +17,10 @@ plugin.tx_seo {
 
 						sortField = datetime
 						lastModifiedField = tstamp
-						pid = {$plugin.tx_news.settings.sitemap.startingpoint}
-						recursive = {$plugin.tx_news.settings.sitemap.recursive}
+						pid = {$plugin.tx_news.settings.sitemap.startingpoint ?? $news.seo.sitemap.startingpoint}
+						recursive = {$plugin.tx_news.settings.sitemap.recursive ?? news.seo.sitemap.recursive}
```
chrissonntag, line 21: Hi there :) I'm just wondering what syntax is actual valid?
`?? $news...` or `?? news...`
With or without $ on the right right of the ??-operator. Or are both valid?

chesio: The variant with `$` on the right side of `??` operator seems to be the right way: https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/Configuration/TypoScript/Syntax/Operators/Index.html#null-coalescing-operator-for-typoscript-constants

chrissonntag: Thanks for clarifying

Issue #2530, opened 2024-09-23
Call to addCacheTags() is deprecated


heinrob:
## Bug Report

**Current Behavior**
The frontend throws this error on the page containing the news list: 

```
TypeError
TYPO3\CMS\Core\Cache\CacheDataCollector::addCacheTags(): Argument #1 must be of type TYPO3\CMS\Core\Cache\CacheTag, array given, called in /var/www/html/vendor/typo3/cms-frontend/Classes/Controller/TypoScriptFrontendController.php on line 424
[...]
at TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController->addCacheTags()
in /var/www/html/vendor/georgringer/news/Classes/Controller/NewsController.php line 92
```
This is a result from this [deprecation: #102422](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.3/Deprecation-102422-TypoScriptFrontendController-addCacheTags.html)

**Expected behavior/output**
The frontend should not throw this error but add the correct `CacheTags` to the respective collector.

**Environment**
- TYPO3 version: 13.3.0-dev
- news version: dev-12-13
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: OSX 14.6.1

**Possible Solution**
Replace calls to `addCacheTags()` with the suggested approach from the deprecation notice.

**Additional context**
A pull request with a potential fix will follow shortly.

PR #2531, opened 2024-09-23
[BUGFIX] Replace deprecated addCacheTags calls

heinrob:
TSFE->addCacheTags() has been deprecated in TYPO3 13.3, see: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.3/Deprecation-102422-TypoScriptFrontendController-addCacheTags.html This results in a TypeError in the frontend where the plugin is used. In this patch, the calls have been updated to use newly instantiated CacheTag objects instead of an array of strings.

Resolves: #2530

### DIFF
--- a/Classes/Utility/Cache.php
+++ b/Classes/Utility/Cache.php
@@ -13,6 +13,7 @@ use GeorgRinger\News\Domain\Model\Dto\NewsDemand;
 use GeorgRinger\News\Domain\Model\News;
 use GeorgRinger\News\Event\ModifyCacheTagsFromDemandEvent;
 use GeorgRinger\News\Event\ModifyCacheTagsFromNewsEvent;
+use TYPO3\CMS\Core\Cache\CacheTag;
 use TYPO3\CMS\Core\EventDispatcher\EventDispatcher;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Extbase\Persistence\Generic\QueryResult;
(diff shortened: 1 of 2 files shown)


Classes/Utility/Cache.php
```diff
@@ -59,6 +60,7 @@ public function isContentRecordAlreadyProcessed(ContentObjectRenderer $cObj): bo
     public static function addCacheTagsByNewsRecords($newsRecords): void
     {
         $cacheTags = [];
+        $cacheDataCollector = $GLOBALS['TYPO3_REQUEST']->getAttribute('frontend.cache.collector');
```
DavidBruchmann, line 63:
```suggestion
        $cacheDataCollector = self::getRequest()->getAttribute('frontend.cache.collector');
```

Classes/Utility/Cache.php
```diff
@@ -83,6 +82,7 @@ public static function addCacheTagsByNewsRecords($newsRecords): void
     public static function addPageCacheTagsByDemandObject(NewsDemand $demand): void
     {
         $cacheTags = [];
+        $cacheDataCollector = $GLOBALS['TYPO3_REQUEST']->getAttribute('frontend.cache.collector');
```
DavidBruchmann, line 85:
```suggestion
        $cacheDataCollector = self::getRequest()->getAttribute('frontend.cache.collector');
```
And add static method to get the request:
```
protected static function getRequest()
{
    return $GLOBALS['TYPO3_REQUEST'];
}
```

georgringer:
thanks for the patch. as v12 is still supported, I need to handle that a bit different
https://github.com/georgringer/news/pull/2544

PR #2584, opened 2024-11-26
[BUGFIX] Replace indexed array keys for items by label and value tags

### LINKED ISSUE #2579: Flexforms are not compatible with changelog from TYPO3 12.3 Deprecation #99739
In commit 768112ec0c, Flexforms got migrated with `<numIndex index="0">` and `<numIndex index="1">` getting respectively changed to `<numIndex index="label">` and `<numIndex index="value">`.

But according to https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.3/Deprecation-99739-IndexedArrayKeysForTCAItems.html, it should have been changed to `<label>` and `<value>` tags.

Also, seems indexes should always be numbers and not strings.

wtfred:
Fix #2579

### DIFF
--- a/Configuration/FlexForms/flexform_category_list.xml
+++ b/Configuration/FlexForms/flexform_category_list.xml
@@ -62,32 +62,32 @@
 								<renderType>selectSingle</renderType>
 								<items type="array">
 									<numIndex index="1" type="array">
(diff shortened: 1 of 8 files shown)


georgringer:
@wtfred please just change the related stuff and not just the full file!

PR #2587, opened 2024-11-28
[TASK] Normalize composer.json

simonschaufi:
The following tool has been used to order all sections in the composer.json file:

https://github.com/ergebnis/composer-normalize


chesio:
It's perhaps good occasion to also add automatic [sorting of packages](https://getcomposer.org/doc/06-config.md#sort-packages):
```json
{
    "config": {
        "sort-packages": true
    }
}
```

PR #2622, opened 2025-02-25
#2621 respect starttime/endtime for cache lifetime of news records

### LINKED ISSUE #2621: Pass a starttime/endtime-based lifetime to the new CacheTag API
You already switched to the `CacheTag` API for core version 13, in `\GeorgRinger\News\Utility\Cache`.

This API supports passing lifetimes per tag. You can pass, pseudo-code!, `($news->getStartime() ?: $news->getEndtime()) - $now` (only if any is set) to respect the start and end times of news records on-the-fly. Then all news lists and detail pages should automatically respect these enabled fields implicitly without having to deal with TypoScript `cache` settings or implementing custom solutions.

ghost:
Resolves #2621 

This passes the starttime and endtime to the CacheTag API if on v13.


stepo2:
Of course it must be per news (also in the demand cache). because every single news (also in a demand) can have an individual start/endtime. `addCacheTagsByNewsRecords` seems only to be for single/detail views.

A cached list view, coming from a demand, containing a news with a starttime would not show the news after this start timestamp without an explicit cache bust. And a listed news with an endtime would be listed after the endtime without an explicit cache bust. That's why _escpecially_ the demand-based cache must use the minimum cache lifetime of all news inside the demand result (and for this it must be iterated over all demand-fetched news to determine the minimum).

I tried it with `addCacheTagsByNewsRecords`, this does not affect any list view at all. The PR can be adjusted to match the latest v13/v14 releases (without the version checks), but formally it is correct otherwise.

PR #2656, opened 2025-07-01
[BUGFIX] Reset cacheTagsToFlush array after flush operation

liayn:


liayn:
@georgringer phpstan update.. not related to this change.

Issue #2666, opened 2025-07-30
Use timestamp including timezone for datePublished meta information


rr-it:
Use timestamp including timezone for datePublished meta information:
See https://developers.google.com/search/docs/appearance/structured-data/article?hl=en#article-types

> ### datePublished 	
> 
> [DateTime](https://schema.org/DateTime)
> 
> The date and time the article was first published, in [ISO 8601 format](https://wikipedia.org/wiki/ISO_8601). We recommend that you provide timezone information; otherwise, we will default to [the timezone used by Googlebot](https://developers.google.com/search/docs/crawling-indexing/googlebot#timezone).
> 
> Add the datePublished property if you want to provide more accurate date information to Google. The [Rich Results Test](https://search.google.com/test/rich-results) doesn't show a warning for this property, as it's only recommended if you decide that it's applicable to your site.

TYPO3 does a nice job to correctly set the timezone for dates now, so we can implement this.

Related to https://github.com/georgringer/news/issues/503

PR #2667, opened 2025-07-30
[BUGFIX] Output timestamp including timezone for datePublished meta data

rr-it:
This fixes https://github.com/georgringer/news/issues/2666


chesio:
I would suggest small change to make the code more readable: `format:'Y-m-d\TH:i:sO'` -> `format:'c'`

The `c` in format string should render ISO 8601 date - see the "Full Date/Time" section here: https://www.php.net/manual/en/datetime.format.php

PR #2672, opened 2025-08-14
[BUGFIX] Fixes required usage of optional parameters

JanStorm:
The three optional arguments imgClass videoClass and audioClass are optional.
Because of that, they can be null when passing to htmlspecialchars.
htmlspecialchars however requires the first argument to be a string, not null and will throw an exception.


georgringer:
thanks, fixed it with 2cb2d222cc89562aea770f3832deb5d92311ec08 a bit different

PR #2725, opened 2025-11-06
Fix `fill` & cleanup BackEnd-Icons

xerc:
w/ `fill`
Bildschirmfoto 2025-11-06 um 16 59 53


xerc:
> Hi, I just happened to come across this PR. In my opinion, it would make sense to also save the icons `Icons/plugin-*` with presentation attributes here in order to remove the inline CSS. This would directly solve the problem described in #2767. What do you think?

not fully ; (vite) build scripts & [EXT:sourceopt](https://github.com/lochmueller/sourceopt/blob/76d4c1cfe27eb0fe101b3c6be14edcaa46f459f2/Classes/Service/SvgStoreService.php#L130-L132) filter `<style|defs>` caused by ..
ref. https://stackoverflow.com/questions/39583880/external-svg-fails-to-apply-internal-css
ref. https://issues.chromium.org/issues/41337331#comment19 ([might be fixed](https://github.com/chromium/chromium/commit/4f3ebd119d5c8492fa8564d61d8b600a82e4a041))

lauralang:
Using the export option Presentation Attributes would not include style tags.

Resources/Public/Icons/plugin-date-menu.svg exported with Presentation Attributes would look like this:
`<svg xmlns="http://www.w3.org/2000/svg" width="96" height="96"><path fill="#fff" d="M0 0h96v96H0z"/><path fill="#7d83de" d="M0 0v96h96V0H0Zm90 90H6V6h84v84Z"/><path fill="#e84f35" d="M33 39h6v6h-6z"/><path fill="#7d83de" d="M21 33v30c0 3.6 2.4 6 6 6h42c3.6 0 6-2.4 6-6V33c0-3.6-2.4-6-6-6H27c-3.6 0-6 2.4-6 6Zm48 30H27V33h42v30Z"/><path fill="#7d83de" d="M45 39h6v6h-6zM57 39h6v6h-6zM33 51h6v6h-6zM45 51h6v6h-6zM57 51h6v6h-6zM39 27h-6v-4.8c0-.6.6-1.2 1.2-1.2h3.6c.6 0 1.2.6 1.2 1.2V27ZM63 27h-6v-4.8c0-.6.6-1.2 1.2-1.2h3.6c.6 0 1.2.6 1.2 1.2V27Z"/></svg>`

georgringer:
@xerc I can't really answer that tbh :( we can also discuss on slack

PR #2726, opened 2025-11-07
Use null coalescing operator for listPid assignment

mxsteini:


georgringer:
but if set to 0, it will now use 0 instead of the id. is that really planned?

PR #2735, opened 2025-11-14
[FEATURE] Add support for PHP 8.5

oliverklee:
Also raise the minimum required TYPO3 bugfix version as needed to avoid deprecation warnings.


oliverklee:
I have no idea why the V13/8.3 builds are failing.

@georgringer Can you help?

DavidBruchmann:
Question is why it passes in tests with other PHP versions.

PR #2737, opened 2025-11-14
[BUGFIX] Drop unused import

oliverklee:


DavidBruchmann:
still fails, but that's perhaps really some hiccup, not sure:
```
 Caused by
mysqli_sql_exception: php_network_getaddresses: getaddrinfo for mysql-func-288 failed: Name does not resolve
```

PR #2754, opened 2025-12-03
Prevent undefined key access

mxsteini:
This will always produce a warning unless you set listPid. The fallback will never be reached if warnings are activated.


georgringer:
fixed with f7f053247569145bf5cb99db7330727c1966b3ce

PR #2794, opened 2026-04-22
[BUGFIX] Remove usage of $GLOBALS['TSFE']

codegm:


codegm:
I would think, that these rector failures have nothing todo with my change

georgringer (approved):

Issue #2774, opened 2026-02-17
Undefined method exception SearchFormViewHelper::registerTagAttribute()


Ekkiheiti:
## Bug Report

**Current Behavior**
If I open a site containing the plugins _News search form_ and _News search result_ TYPO3 throws the following exception:

_Call to undefined method GeorgRinger\News\ViewHelpers\SearchFormViewHelper::registerTagAttribute()_

**Expected behavior/output**
Loading my site showing the search form.

**Environment**
- TYPO3 version(s): 14.1.0 with Fluid 5.1.0
- news version: 14.0.1
- Is your TYPO3 installation set up with Composer (Composer Mode): yes
- OS: Debian 12

**Possible Solution**
According to [TYPO3's changelog](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.3/Deprecation-104223-FluidStandaloneMethods.html) the calls of `$this->registerTagAttribute` inside `SearchFormViewHelper::initializeArguments()` aren't working since Fluid v5. Maybe they should converted to `registerArgument()` calls instead?

euwa:
Hello, is there any further information on this topic yet, or is there an estimate of when a fix will be available? Many thanks for your work on this extension.

DavidBruchmann:
I'm just working it.
A PR is available still today, on merging or releases I've no influence.

DavidBruchmann:
So, the PR is available now!

For fixing this and perhaps further issues, you can [Support me on PayPal](https://paypal.me/SophieBarlian) (it's my wife's account).  
We need urgently a bit, 5 or 10 € help already!

PR #2804, opened 2026-05-27
Update argument registration and request handling for TYPO3 14

spfeifer:
- registerTagAttribute usages have been removed where possible and changed to registerArgument where necessary
- Server request handling was updated to work with TYPO3 14


georgringer:
fixed now with #2774

PR #2818, opened 2026-07-27
[TASK] Resolve category subtrees without one query per category

bmack:
## Problem

`CategoryService::getChildrenCategories()` walks the category tree with one query per category:

```php
$res = $queryBuilder->select('uid')->from('sys_category')
    ->where($queryBuilder->expr()->in('parent', ...))
    ->executeQuery();

while ($row = $res->fetchAssociative()) {
    $subcategories = self::getChildrenCategoriesRecursive((string)$row['uid'], $counter);
    ...
}
```

On an installation with 3152 categories, resolving all root categories takes **2926 queries and about a second**. It is cached in `news_category`, but the cost is paid again after every cache flush, for every id list variant. I found this while profiling a news page: the category menu spent ~1s of a 4s uncached render inside this method.

## Change

The categories are now collected level by level, so the number of queries depends on the **depth** of the tree instead of on the **number** of categories. The traversal itself then happens in PHP on the collected `parent => children` map, still depth first, so every category is directly followed by its own descendants.

## Result

Measured on the installation mentioned above (3152 categories, 227 root categories):

| id list | queries before | queries after |
|---|---|---|
| all 227 root categories | 2926 | 3 |
| 5 root categories | 2507 | 3 |
| single category | 1-269 | 1-2 |

Over a set of 45 different id lists: **15838 queries before, 70 after**, with identical id sets returned in every case.

## Behaviour

I compared the old and the new implementation over 45 id lists against the real 3152 category table. The returned id sets are identical. Two differences worth mentioning:

* A category that is its own ancestor no longer produces 10000 entries before the recursion guard stops the traversal. The new test `getChildrenCategoriesTerminatesOnCyclicCategories` fails on `main` (it returns 10000 times the same uid) and passes with this change.
* A category that is both passed in **and** a child of another passed in category is no longer contained twice in the result. `getChildrenCategories('15,16,17')` where 17 is a child of 16 returned `15,16,17,17,18,...` before and returns `15,16,17,18,...` now.
* The order of the flattened list groups the descendants per passed in category, instead of interleaving them in the order the database happened to return the first level. The query never had an `ORDER BY`, and both consumers (`IN (...)` constraints and `CategoryRepository::findTree()`, which re-sorts by `sorting`) do not depend on the order.

## Tests

Added `Tests/Functional/Service/CategoryServiceTest.php` with a nested category fixture covering nested subtrees, multiple start categories, hidden and deleted categories, categories without children, unknown categories, `$removeGivenIdListFromResult` and the cyclic case. 9 of the 10 tests pass on `main` as well, so they double as characterisation tests for the unchanged behaviour.


bmack:
Note on the red CI: the failing jobs are the two pre-existing `PluginPreviewRendererTest` errors (`Typed property PluginPreviewRenderer::$iconFactory must not be accessed before initialization`), which also fail on `main` and are unrelated to this change. The unit suite reports the same `Tests: 253, Errors: 2` as the `main` baseline.

Because the unit step exits first, the functional step never runs in CI, so the added `Tests/Functional/Service/CategoryServiceTest.php` is not covered by the pipeline yet. I ran the full functional suite locally against MariaDB 11.8 on PHP 8.5: `OK (30 tests, 67 assertions)`.

PR #2819, opened 2026-07-27
[TASK] Do not load every parent category to build the category tree

bmack:
## Problem

`CategoryRepository::findTree()` resolves the parent of every category through the getter:

```php
foreach ($categories as $category) {
    $flatCategories[$category->getUid()] = [
        'item' => $category,
        'parent' => ($category->getParentcategory()) ? $category->getParentcategory()->getUid() : null,
    ];
}
```

`parentcategory` is a `#[Lazy]` relation, so `getParentcategory()` calls `_loadRealInstance()`, which runs a query plus the language overlay for **every single category in the tree** — only to read the uid from the loaded object. The parent objects themselves are never used here.

I found this while profiling a news page on an installation with a few thousand categories: rendering the category menu spent a noticeable part of its time in `Typo3DbBackend::getObjectDataByQuery()` and `PageRepository::getRecordOverlay()`, all triggered from this loop.

## Change

`Category::getParentcategoryUid()` reads the uid from the lazy loading proxy without resolving it, and `findTree()` uses that instead. The new method is also useful on its own — reading the uid of a lazy relation without fetching it is a common need.

```php
public function getParentcategoryUid(): ?int
{
    $parentcategory = $this->parentcategory;
    if ($parentcategory instanceof LazyLoadingProxy) {
        return $parentcategory->getUid() ?: null;
    }

    return $parentcategory?->getUid() ?: null;
}
```

## Result

On a news page with 374 categories in the tree, **326 of them were behind an unresolved lazy loading proxy**, so this saves 326 queries (plus their language overlays) per `findTree()` call.

## Correctness

The uid on the proxy is the raw `parent` value of the row Extbase loaded, while the old code read the uid of the mapped parent object. I checked that these agree on an installation with 3152 categories, of which 3010 are translations, comparing both values for every category in three site languages: **374 categories checked, 0 differences.**

## Tests

* Unit tests for `getParentcategoryUid()`, including one asserting that `DataMapper::fetchRelated()` is **never** called, which is the actual point of the change.
* Functional tests for `findTree()` covering nested categories and the "leaf selected without its parent" case. `findTree()` had no test coverage so far.

## Note on the null guard

The added `findTree()` tests fail on `main` under PHP 8.5 with *"Using null as an array offset is deprecated"* (the suites run with `failOnDeprecation="true"`), because a root category has no parent. This PR therefore contains the same one line guard as #2788. If #2788 is merged first, this part becomes a no-op and I am happy to rebase.

### DIFF
--- a/Classes/Domain/Model/Category.php
+++ b/Classes/Domain/Model/Category.php
@@ -211,6 +211,22 @@ class Category extends AbstractEntity
             : $this->parentcategory;
     }
 
+    /**
+     * Uid of the parent category, without loading the parent category itself.
+     *
+     * Use this instead of getParentcategory()->getUid() if only the uid is needed,
+     * as the parent category is a lazy relation and would have to be fetched first.
+     */
+    public function getParentcategoryUid(): ?int
+    {
+        $parentcategory = $this->parentcategory;
+        if ($parentcategory instanceof LazyLoadingProxy) {
+            return $parentcategory->getUid() ?: null;
+        }
+
+        return $parentcategory?->getUid() ?: null;
+    }
+
     public function setParentcategory(Category $category): void
     {
         $this->parentcategory = $category;
--- a/Classes/Domain/Repository/CategoryRepository.php
+++ b/Classes/Domain/Repository/CategoryRepository.php
@@ -104,7 +104,7 @@ class CategoryRepository extends AbstractDemandedRepository
         foreach ($categories as $category) {
             $flatCategories[$category->getUid()] = [
                 'item' => $category,
-                'parent' => ($category->getParentcategory()) ? $category->getParentcategory()->getUid() : null,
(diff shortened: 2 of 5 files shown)


bmack:
Note on the red CI: the failing jobs are the two pre-existing `PluginPreviewRendererTest` errors (`Typed property PluginPreviewRenderer::$iconFactory must not be accessed before initialization`), which also fail on `main` and are unrelated to this change.

The four added unit tests do run and pass in CI: the unit suite reports `Tests: 257` here against `Tests: 253` on `main`, with the same two pre-existing errors.

Because the unit step exits first, the functional step never runs in CI. I ran the full functional suite locally against MariaDB 11.8 on PHP 8.5: `OK (30 tests, 67 assertions)`.

